PR #47, opened 2023-05-10
Update LICENSE

rewtd:


hblankenship (approved):
So under both the CC and the Apache 2.0 (I assume for the build code).

rewtd:
Indeed!

PR #49, opened 2023-05-18
Fixes for hashed files.

rewtd:


hblankenship (approved):
Ah, that supply chain....

PR #63, opened 2023-05-24
CRE Mappings

northdpole:
hi Cornucopia friends, here's the mappings we've been discussing about and a script to make future mappings easier to get.

Keep in mind the script only creates a new mapping suggestion, manual intervention is needed in order to validate mapping correctness


source/add_cre.py
rewtd, line 1: Can this file be moved into the /scripts folder please? That way the code to generate the cards is all in the same place. Once in that folder however, it will need to have some unit tests added.

source/add_cre.py
```diff
@@ -0,0 +1,100 @@
...
+                pprint(
+                    f"for manual mapping card is {card_val_to_desc[name][card.get('value')]}"
+                )
+                pprint("-------------------")
+
+            my_set = set(itertools.chain.from_iterable(res2))
+            result[index]["suits"][suitsIndex]["cards"][cardIndex]["cre"] = list(my_set)
+            result[index]["suits"][suitsIndex]["cards"][cardIndex]["description"] = card_val_to_desc[name][card.get("value")]
+
+with open("./cre-mappings-1.2.yaml", "w") as outf:
```
rewtd: Eventually we will want this script to generate the /source/ecommerce-mappings-1.2.yaml file directly, you say there's a manual verification required however, how is that done?

source/cre-mappings.yaml
```diff
@@ -0,0 +1,725 @@
...
+meta: {component: mappings, edition: ecommerce, language: ALL, version: '1.20'}
+suits:
+- cards:
+  - capec: [54, 541]
+    cre: [308-515]
+    description: Brian can gather information about the underlying configurations,
+      schemas, logic, code, software, services and infrastructure due to the content
+      of error messages, or poor configuration, or the presence of default installation
+      files or old, test, backup or copies of resources, or exposure of source code
+    owasp_appsensor: [HT1-3]
```
rewtd: These descriptions are in the cards files so that translation is easier, they should not be included in the mappings files as they'll not get used to generate the cards, unless they are needed to create the CRE mapping?

source/cre-mappings.yaml
```diff
@@ -0,0 +1,725 @@
...
+    owasp_asvs: [1.1, 4.5, 8.1, 11.5, 19.1, 19.5]
```
rewtd: Are these generated from CRE or pulled from the existing mapping file? If the former, that's brilliant - if the latter then they are wrong and we'd be better off generating them from the CRE mapping, is that possible?

northdpole: they were supposed to be from the existing mapping file. I'll generate them from the website then

source/cre-mappings.yaml
northdpole: i was using them as a visual queue to double check the mapping is correct, now that we have the mappings I'll remove them, thanks!

source/cre-mappings.yaml
rewtd: I see that you mapped the Open CRE IDs from the CAPEC IDs and that's a great choice. If we can then use the OCRE IDs to get the ASVS ones that would be amazing!

northdpole:
Did several changes and even added unit tests.
Details:
* source now has a file mapping only cornucopia to CREs this can be used as your "core" so that you can populate more mappings from this base
* scripts has a gen_mappings file which can be run as a library or by itself, it has two methods, one to enhance your base mappings with any other information that opencre.org knows about and another one to generate qr codes for each cre
* added a unit test

Caveats:
* The mappings are manual, and I very much suggest you confirm they are what you need or you update them.
* Opencre does not know of appsensor, safecode or scp yet, we will add them soon however, but perhaps you want to choose some other sources from staging or opencre.org

let me know what else you need!

rewtd (approved):
LGTM

PR #106, opened 2023-06-22
Create CODE_OF_CONDUCT.md

rewtd:


hblankenship (approved):
Not opposed to having these per-repository but it does make me wonder why there isn't a simpler way to have the ORG host files like this for all repos under the org; that way when it gets updated, it is in one place....

rewtd:
100% I have literally just copied the Code of Conduct from the OWASP
website here...

Issue #13, opened 2022-10-19
ASVS v4


sydseter:
Any interest out there for creating a version of the game for ASVS 4.0?
We are using the game internally to be able to help in defining security constraints for our epics and user-stories ([ASVS v4, 1.1.3](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x10-V1-Architecture.md#v11-secure-software-development-lifecycle)), but I find that I end up translating the ASVS v3 requirements into v4. I am realizing that creating additional cards for a ASVS 4.0 version would be the best.

rewtd:
Hi @sydseter absolutely! Or creating a new ecommerce-mappings-2.0.yaml file with the mappings corrected on it? Once I get the test framework working again I'll jump into this.

sydseter:
Great, you might want to consider exchanging some scenarios for new cards and perhaps even renaming or changing categories as well, but I can start with translating the easy parts first. Perhaps the rest can be discussed?

sydseter:
[OWASP ASVS 3.0 & 4.0 Comparison mapped to Cornucopia.xlsx](https://github.com/OWASP/cornucopia/files/13887412/OWASP.ASVS.3.0.4.0.Comparison.mapped.to.Cornucopia.xlsx)

Hi, I have created a mapping for ASVS 3.0.1 to ASVS 4.0 that can be used with Cornucopia.

I have also finished implementing changes to the cli application so that it is possible to convert between different mappings versions so that it is possible to go back and publish updates to Cornucopia 1.21 ecommerce version if needed. I am doing the final testing and will create a pull-request momentarily. For now, please, if someone can look at the ASVS 3.0.1 to 4.0 mapping and comment where needed, then I can make some last minute adjustments. 

Should I add the idml, docx and pdf files to the output folder so that they can be reviewed?

I found several copy&past errors in the translations. I have fixed these as well.

sydseter:
I could also give a hand in updating https://copi.securedelivery.io/ as well if you trust me and need a hand. We use it on a regular basis in our company.
I have had some issues getting others to facilitate the copi sessions. 
Updating the ASVS mapping to 4.0 would be a big help.

sydseter:
I realized now that there is an error in the excel here, but it didn't spill over in ecommerce-mappings-1.3.yaml so I guess it's fine. 

Just want to point it out in case you decide to have a second look.

The error is found in the Excel under B7. The right value is:

[...]

PR #303, opened 2024-01-22
Translation adaptations and completions for multiple languages and link corrections to ensure the translations are compatible with both ASVS v4 and v3

sydseter:
## This pull request contains

- Multi-language fixes for #13
- Adapt the static and dynamic templates so that they can be printed correctly in multiple languages.
- Complete the Spanish language translations.
- Changed the docx template margins for the cards.
- Fix copy&past errors in the Spanish translation.
- Removed wrongly formated text that was included in the translations.
- Added generated output after applying the changes.

### Multilanguage templated adaptations
Previously the translated docx versions had issues because the docx templates weren't really adapted to work with multiple languages. I therefor added new table rows and moved the text from the last table to the first table in the docx templates so that the page breaks comes where you would expect them. This means that you know can add more languages without having to worry about the docx templates. The previous versions would also have empty space spill over on the next page. In order to fix this, the column height has been adjusted and white space removed so that the text doesn't spill.

#### **Previously:**

issues with page breaks 2

#### **Now:**

issues with page breaks 2 fixed

#### **Text and white space spilling issues has now been fixed**

**ex. 1:**
empty space

**ex. 2:**
issues with page breaks

### ASVS v4 fixes for links
Previously the link references where quite long. This would break the docx templates. In order to make the templates work with multiple languages, the links had to be shorten to not break the columns in the table. Some of the links also referred only to ASVS v3. In order for the templates to be backward compatible the references no points to the OWAS ASVS repository where both the v3 and the v4 printable documents can be downloaded.

#### **Link Examples**
 
**previously ex. 1:**
issues with links

**now ex. 1:**
issues with links fixed

**previously ex. 2:**
old links

**now ex. 2:**
old links fixed

### Removal of included bullet points and tab spaces in the translation that shouldn't have been included.

Some of the translations has the bullet points included. 
The French translation has several tabs that should have been a single space instead.
This has been cleaned up.


sydseter:
I am not sure that test will work since there is no reference to the branch during a pull-request event. Tried to change it, but I did update all the output files at the end here, so they should be ok.

sydseter:
You mentioned that the bulletpoints might have been used in the leaflet, although that might be true, there is no automatic generation of the leaflet, there is just an original that always have been manually edited. see: https://github.com/OWASP/cornucopia/blob/master/resources/originals/OWASP%20CORNUCOPIA%20Folded%20Leaflet%20v1.20-EN.idml
Even if we where to create an automatic generation of the leaflet, I think we still should opt for not having the formatting like bullet-points in the translations. Reason being that it cause issues for how the text breaks and flows for the various templates. ex, In the docx you end up with double bullet, points or whitespace that breaks the translated sentences in places you wouldn't want it to. e.g:

ex 1:
white space issues2

ex 2:
white space issues


 I think it would be better to edit the templates and add bullet points there when that is needed. Keeping the formatting and links in the templates and the translated content in the yaml files will make it much easier to add new translations.
 Otherwise you will end up adapting the templates when adding new translations.

sydseter:
Here are the same examples as they look like after I cleaned them up:

ex 1:
white space issues2 fix


ex 2:
white space issues fix

sydseter:
I understand that we should maybe keep the cleanup in a separate pr, but it's better to have all of it here, since what we care the most about here, is to check that the templates works with **all** of the translations, this way, it makes it easier to see that the output, does in deed works across all of the printable files. I couldn't just do the fixes for ASVS v4 since the result wouldn't be printable.

sydseter:
I may have gone a bit overboard changing AZ3 description for the english translation using "he" instead of "they". "they" is of course better.

rewtd:
> You mentioned that the bulletpoints might have been used in the leaflet, although that might be true, there is no automatic generation of the leaflet, there is just an original that always have been manually edited. see: https://github.com/OWASP/cornucopia/blob/master/resources/originals/OWASP%20CORNUCOPIA%20Folded%20Leaflet%20v1.20-EN.idml 

There should have been an automated generation of the leaflet, although I didn't do the work on that. That original file was indeed manually created. I agree with the rest of what you wrote however.

rewtd (approved):
Lots to look at here and it's probably good to have this in the main branch anyhow.

Issue #288, opened 2024-01-12
Missing eps graphics for the back of the cards.


sydseter:
**Describe the bug**
I started reviewing the idml files in Indesign and noticed that there is a hardcoded link to a eps file which provides the background for the card. I can't find it in the repository which lead me to believe that it is missing and hasn't been provided in the original template. 

We should find a way to include the graphics in a way so that it is either embedded or properly linked to during the idml conversion.
please have a look at the screenshot from Indesign with the error and the pdf that I printed after opening the file in Indesign.

**To Reproduce**
1. Open any of the idml files in Indesign and try to print the deck
2. Choose Windows -> Links to get the list of errors

**Expected behavior**
The graphics should not be missing if they are needed in order to print the cards.

**Screenshots**
![backofcards_eps_missing](https://github.com/OWASP/cornucopia/assets/1244005/14309a59-2f6c-4c69-a84b-5520b8d267b2)

**Error:**
Name	cornocopia back of card elements.eps
Format	EPS
Page	1
Status	Missing
Size	2.8 MB (2896558 bytes)
Title	Print
Creator	Adobe Illustrator CS6 (Macintosh)
Place Date	Monday, July 11, 2016 1:38 PM
Layer	Default
Path	\Users\agb\Downloads\cornucopia-ecom-1v20en-owasp\Cards\Links\cornocopia back of card elements.eps
Creation Date	Thursday, October 10, 2013 5:22 PM
Scale	100%

**Desktop (please complete the following information):**
 - OS: Windows
 - Version Cornucopia ecommerce edition 1.30

**Additional context**
Pdf output from Indesign: [cornucopia.pdf](https://github.com/OWASP/cornucopia/files/13918303/cornucopia.pdf)

rewtd:
Let's catch up on this? Are you on Slack?

sydseter:
Just ping me when you have the time.

sydseter:
@rewtd I am able to embed the eps file into the idml, but it will increase the filesize with 4MB which is quite a lot. I am thinking we could remedy that by using https://git-lfs.com/ 
I think we will anyway need to find a solution like this since we have opted for storing the output files. Over time, it will become impossible to clone the repository due to the size of the index that need to keep track of changes between the each version of binary files. This will solve that.

Please tell me if you have any concerns about this.

PR #313, opened 2024-01-25
#288 Embed missing background file into the InDesign templates

sydseter:
In this pull-request:

- Embed missing background file into the InDesign templates
- Make sure the updateoutputfiles github action builds and commits the files correctly to the pull-request
- Add git-lfs


rewtd (approved):
It's difficult to review the templates but this looks good to me! Loving the move to lfs too!

PR #328, opened 2024-02-02
Fixing dead links.

sydseter:
In this pull-request:

- Links in static and dynamic docx template has been fixed

The OWASP SCP, SAFECode links where dead and the CAPEC was pointing towards an old xml file.
The links has been renewed and updated.

Before:
broken links

Now:
Fixing the links in the docx template


rewtd (approved):
This is good!

PR #329, opened 2024-02-02
Fixes #289 Exchanging fonts restricted for commercial use with font licensed under SIL Open Font License - OFL (Free for Commercial Use)

### LINKED ISSUE #289: Apple Braille Outline 6 Dot doesn't have any suitable alternative on Windows
**Describe the bug**
Apple Braille Outline 6 Dot is one of the fonts used in the idml file. There are no suitable alternative to this font on any windows system. To simplify the printing process. We should use fonts that can be installed on both systems when needed. The Myriad Pro Semibold font is also not a font that is available, but it can at least be bought from adobe, still, I would recommend finding a suitable open source substitution for this font as well as it would simplify development and printing.

**To Reproduce**
1. Open any idml file in Indesign on a windows system
2. Missing Fonts will get reported

**Expected behavior**
Ideally we should choose fonts that are easy to require on both windows and mac os x

**Screenshots**
font missing

**Desktop (please complete the following information):**
 - OS: Windows 10

sydseter:
This pull-request contains:

- Fixes #289
- Apple Braille Outline 6 Dot has been replaced with **Atkinson Hyperlegible**: https://brailleinstitute.org/freefont
- Myriad and Myriad Pro has been replaced by Fivo Sans: https://www.fontsc.com/font/fivo-sans

The new fonts can be used on any platform and computer software without having to pay license fees.

The fonts are slightly different then the old fonts, but they should provide better readability and I believe they are prettier as well. 

Here is the finished result:

Before;
[OWASP CORNUCOPIA Playing Cards v1.pdf](https://github.com/OWASP/cornucopia/files/14143787/OWASP.CORNUCOPIA.Playing.Cards.v1.pdf)

Now:
[owasp_cornucopia_ecommerce_cards_nl_1_30.pdf](https://github.com/OWASP/cornucopia/files/14143790/owasp_cornucopia_ecommerce_cards_nl_1_30.pdf)


sydseter:
Information on Fivo Sans: https://www.behance.net/gallery/53653425/Fivo-Sans-Typeface-Free-Font-Family

Information on Atkinson Hyperlegible: https://en.wikipedia.org/wiki/Atkinson_Hyperlegible

sydseter:
They are both «grotesque» fonts which gives them a similar feel: https://fonts.google.com/knowledge/glossary/grotesque_neo_grotesque

I only replaced the font for the «Joker» text with Atkinson as it was the only place Apple braile font was used. The Idea might have been to use a higly readable font for the card numbers, but it doesn’t seem like this was finalized. 

Fivo Sans is great when you want to have a font in multiple languages. It supports 192 languages and comes in multiple variants making it very versitile and multi-purpose.

rewtd (approved):
Much better! Thank you!

PR #392, opened 2024-04-06
Run tests when bumping the dependencies or changing the pipeline in order to get confirmation that everything is ok

sydseter:
It's a "good" practice to run the tests and generate artifacts when patching the dependencies or the pipeline as well. It should make it safer to merge in dependencies and pipeline updates.


sydseter:
Not possible to checkout the pull-request from another repository in our pipeline in order to generate the artifacts and test them. Security wise this is probably for the best.

sydseter:
The artifacts need to be associated with a branch. It can not be master so I will need to test somethings out for this to work.

sydseter:
By commiting the artifact to Git LFS for the pull-request, the artifacts can be displayed. This ofcourse generated quite a large number of commits that then is merged to master. The alternative is to choose a branch, e.g: artifacts where we push the commits.

The alternative is to create a temporary branch, based on the pull-request name, push the artifacts there, then delete the branch after merge,

I can look into the later if you feel that is a better solution.

rewtd:
Either way works for me.

Although: Having a special branch (artefacts) means a dependency that could
break stuff and creating a branch and cleaning up afterwards is more
git-like.

Let’s try the latter first?

rewtd (approved):
You are right, we should do a build on dependency changes. Tests don't cover everything.

PR #393, opened 2024-04-06
Revert "Run tests when bumping the dependencies or changing the pipeline in order to get confirmation that everything is ok"

sydseter:
Reverts OWASP/cornucopia#392

That did not work. Not sure why it is commiting to the base of the repository.


rewtd (approved):
Back to the drawing board, I guess?

Issue #308, opened 2024-01-24
Fuzzer test fails unexpectedly


rewtd:
I'll have a look at the fuzzer test. Something strange going on there.

_Originally posted by @sydseter in https://github.com/OWASP/cornucopia/issues/306#issuecomment-1908031468_

PR #395, opened 2024-04-07
Adding hypothesis test example and start on the first cornucopia fuzzer

sydseter:


sydseter:
This still require a bit more work. Hard to test this out. The tool behaves very differently locally.

sydseter:
I got the fuzzer working and added the first test. We need to improve the coverage as we add more code.
At least it works now.

rewtd (approved):
This is good work!

sydseter:
related to #308

PR #405, opened 2024-04-15
Delete old workflow, test reusable workflow for running tests.

sydseter:
In this pull-request:

- run tests when patching and for pipeline changes, make the workflow for running tests reusable.


sydseter:
I was thinking that it's a bit better if we run the unit tests for each of the github dependency patches and pipeline changes as well so that we have some confidence that they actually work.

I have reused the hardening and run-tests jobs so that we don't duplicate the pipeline jobs this should make it a bit easier to maintain the code.

PR #408, opened 2024-04-17
Adding hardening on run-tests and delete-artifacts in case it is run separatly.

sydseter:


sydseter:
fixes #407

rewtd (approved):
Perfect! Does this negate the other commit?

sydseter:
@rewtd I doubt it, but it is worth a try.

PR #409, opened 2024-04-17
Remove temporary files.

sydseter:


rewtd (approved):
Looks good. It's only a tidy up and the big change is the removed of the checkout ref?

sydseter:
@rewtd correct.

PR #424, opened 2024-04-22
Removing ASVS 6.1.1 from cryptography 4 since this deals with encryption at rest while the card is specified for encryption in transit.

sydseter:
6.1.1 and 7.3.2 has been removed from asvs 4.0.3


rewtd (approved):
LGTM

PR #425, opened 2024-04-22
Updating the dependencies for fuzzing according to recommendations given by the dependabot.

sydseter:
fixes: #421 , #415 and #412

Didn't like the latest update from dependabot, so I have tested them out and edited the update.


rewtd (approved):
LGTM

PR #426, opened 2024-04-22
ASVS 1.4.3 has been considered a duplicate of 4.3.1 in ASVS 4.0.3 and removed. This will reflect that change.

sydseter:
see: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x10-V1-Architecture.md?plain=1#L46


rewtd (approved):
LGTM

PR #436, opened 2024-04-24
Ensure ASVS mapping can be processed programaticly.

sydseter:
In order to be able to create proper links, and process the mapping codes individually, we need to specify each code separately explicitly and not implicitly with an n-dash.


rewtd (approved):
Good idea! Do the strings still fit in the space for them in the templates?

sydseter:
@rewtd Yes the text expands over two lines if necessary thanks to the logic in convert.py. I checked that it’s ok.

PR #446, opened 2024-04-25
Ensure whitespace doesn't get trimmed when the text placeholders are replaced inside the paragraph.

sydseter:
In this pull-request:
- Improve the consistency and readability for the asvs, capec, scp and appsensor code tables 
- Support for having multiple text placeholders within the same paragraph without getting tabs, space and other whitespace characters stripped. Needed for correctly being able to build the leaflet.

I found some complicated code related to the replacing the text placeholders correctly even if the case was wrong. Very considerate, but it added a lot of complexity that didn't really take away the need for doing QA so I removed it. (only used for the idml files btw.)

[owasp_cornucopia_ecommerce_cards_en_1.30_leaflet.pdf](https://github.com/OWASP/cornucopia/files/15123285/owasp_cornucopia_ecommerce_cards_en_1.30_leaflet.pdf)

[owasp_cornucopia_ecommerce_cards_en_1.30_static.pdf](https://github.com/OWASP/cornucopia/files/15123287/owasp_cornucopia_ecommerce_cards_en_1.30_static.pdf)

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -513,14 +513,8 @@ def get_replacement_dict(input_data: Dict[str, Any], mappings: bool = False) -> 
 def get_replacement_mapping_value(k: str, v: str, el_text: str) -> str:
     reg_str: str = "^(OWASP SCP|OWASP ASVS|OWASP AppSensor|CAPEC|SAFECODE)\u2028" + k.replace("$", "\\$").strip() + "$"
     if re.match(reg_str, el_text.strip()):
-        pretext = el_text[: el_text.find("\u2028")]
-        v_new = v
-        if len(v) + len(pretext) > 60:
-            v_new = v.replace(", ", ",")
-        if len(v_new) >= 34:
-            v_split = v_new.find(",", 25 - len(pretext)) + 1
-            el_new = pretext + "    " + v_new[:v_split] + "\u2028" + v_new[v_split:].strip()
-            return el_new
+        if len(v) >= 38:
+            return el_text[: el_text.find("\u2028")] + ": " + v
         else:
             return el_text.replace(k, v)
     return ""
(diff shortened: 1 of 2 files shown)


tests/scripts/convert_utest.py
```diff
@@ -1488,7 +1488,6 @@ def setUp(self) -> None:
         self.replacement_values = [
             ("${VE_suit}", "Validation & Encoding"),
             ("${VE_VE2_desc}", "You have invented a new attack against Data Validation and Encoding"),
-            ("${VE_VE2_misc'}", "Read more about this topic in OWASP's free Cheat Sheets"),
```
rewtd, line 1491: Is the _misc never called?

scripts/convert.py
```diff
@@ -513,44 +513,22 @@ def get_replacement_dict(input_data: Dict[str, Any], mappings: bool = False) ->
...
-        v2: str = v.strip()
-        if el_text == k:
-            return v
-        elif el_text.strip() == k2:
-            return v2
-        elif el_text.lower() == k.lower():
-            return v
-        elif el_text.strip().lower() == k2.lower():
-            return v2
-        elif convert_vars.making_template:
```
rewtd, line 541: This code allowed templates to be created in the first place - it's probably no longer required and it exists in the git history if we ever need it again.

rewtd (approved):
Looks good - assuming _misc is not used for anything?

tests/scripts/convert_utest.py
sydseter: This is not really testing misc it’s testing how templating is suppose to work if the template label has the wrong case, contain apostroph or the template value is surrounded with whitespace. I removed it as this functionality is not used and make the code complicated and error prone.

scripts/convert.py
sydseter: Yes, I would not automaticly build templates. Keep in mind that we are talking about idml templates here, not yaml or docx

PR #451, opened 2024-04-26
Translate from english.

sydseter:
In this pull-request:

- Some headlines I had overlooked

These I had overlooked. 
It would be fantastic if you could look over the Dutch translations and see what we need to correct there.


source/ecommerce-cards-1.30-nl.yaml
```diff
@@ -473,7 +473,7 @@ paragraphs:
     text: "De inhoud is voornamelijk afkomstig uit het SCP."
   -
     value: "T00500"
-    text: "Mappings"
```
rewtd: In Dutch we'd use "mapping" in Dutch too, to talk about the kind of mapping we mean here.

rewtd (approved):
The other change is good however.

PR #459, opened 2024-04-29
Add the cornucopia mobile deck.

sydseter:
In this pull-request:

- This will add the Cornucopia mobile deck to the solution.
- The design is based on the MASVS website where most of the scenarios and requirements come from. see: https://mas.owasp.org/MASVS/
- please see pdf: 
[owasp_cornucopia_masvs_cards_en_1.00_masvs.pdf](https://github.com/OWASP/cornucopia/files/15148770/owasp_cornucopia_masvs_cards_en_1.00_masvs.pdf)


rewtd (approved):
This works for now - but eventually we want to only produce "current" versions I think and probably we want to generate all cards (mobile or otherwise) at the same time.

sydseter:
@rewtd Yes, I was thinking more that we probably need to support multiple versions for multiple editions, but implementing that now require quite a few changes. This code is still sufficient for creating multiple editions and leaflets at the same time.

PR #461, opened 2024-04-30
Correcting the Joker text.

sydseter:


sydseter:
Not even, completely bogus. I should get better at QA.

rewtd:
That is why we work together ;-)

PR #473, opened 2024-05-02
Change the card codes.

sydseter:
Need to change the card codes to create unique urls for copi.


rewtd:
What was the need to update the IDs a result of? This will have an impact (eventually) on the per card URL.

rewtd:
> What was the need to update the IDs a result of? This will have an impact (eventually) on the per card URL.

Never mind. It's because the mobile deck would conflict with the web one.

rewtd (approved):

PR #475, opened 2024-05-02
Update tests to reflect the new mobile suite.

sydseter:
In this pull-request:

I ended up merging in the mobile codes thinking that the tests was ok, turned out the pipeline doesn't run them because of a mistake i made.


rewtd (approved):
LGTM

PR #479, opened 2024-05-03
Correction to credits, and clarification about mappings

cw-owasp:
Mappings relate to Web App Edition, to allow for upcoming Mobile App Edition mappings


rewtd (approved):
Looks good!

PR #489, opened 2024-05-07
Fix licensing.

sydseter:
WIP


sydseter:
Thought I take the initiative here. The current changes that I made is probably wrong. It would be great if you could give me some directions. I understood that part of the content was Apache 2.0 licensed. If so, which part?
Having both the "CC-BY-SA-3.0" and "Apache 2.0" licenses at the base of the repo without any note like now, is not a good idea. If most of the content is licensed under  "CC-BY-SA-3.0" we just need to note where that is not the case as these licenses not is considered to be compatible under the CC-BY-SA-3.0 terms.

README.md
```diff
@@ -8,18 +8,44 @@ OWASP Cornucopia is a mechanism in the form of a card game to assist software de
...
+
+© 2010 Microsoft Corporation. Text for Elevation of Privilege (EoP) is licensed under [CC-BY-SA-3.0](./LICENSE.md) 
+
+### OWASP Cornucopia Mobile App Edition
+
+Text and code mapping for OWASP Cornucopia Mobile App Edition is licensed under [CC-BY-SA-3.0](./LICENSE.md)
+
+### OWASP Cornucopia Website App Edition
+
+Text and code mapping for OWASP Cornucopia Website App Edition is licensed under [Apache License 2.0](./source/LICENSE.md)
```
rewtd: I think the original licence was Apache License 2.0 - but I'd like to make the Website App Edition CC-BY-SA-3.0 same as the rest.

rewtd (changes requested):

README.md
rewtd: I've updated things to reflect that. But... have I done it right?

sydseter: Great! Looks good to me.

sydseter: @rewtd since I opened the pull-request, you need to approve it.

rewtd (approved):

PR #494, opened 2024-05-08
Pin version.

sydseter:


sydseter:
We'll just remove this script as it isn't necessary anymore.

rewtd (approved):

PR #507, opened 2024-05-11
Fixup attribution on the case

sydseter:
Ensure it says: "based on an idea by Colin Watson", on the mobile case.


cw-owasp:
Agreed.

cw-owasp (approved):

PR #528, opened 2024-05-16
Prepare for 2.0 release and mobile app release and shorten the build time.

sydseter:
Instead of building everything, let's just build what we need to build and test.


rewtd (changes requested):
Perhaps we should also remove the "dynamic" from the -all list for now?

sydseter:
Yeah that's a good idea.

sydseter:
done!

rewtd (approved):
LGTM - we do need copi up on owasp.org for this to work however...

sydseter:
@rewtd This may not be the right way to generate the qr codes. They need to be embedded into the idml and docx files and indesign already have a way of generating qr codes. I am wondering whether we should remove it.

rewtd:
100%

PR #558, opened 2024-05-24
Add credits to secure delivery for copi

sydseter:


cw-owasp:
These all look good, thank you.

PR #560, opened 2024-05-25
Cornucopia 2.0

sydseter:
As we are releasing the 2.0 version, I would like to do some improvements that might not be backward compatible and that I have been wanting to do.
In order to make it easier to add more deck, we should remove some of the old versions that we are not using any more.
I think we can remove the templating function as well as it's basically copying from an old template which you will have to edit manually anyway. It makes the code harder to read and understand without offering much benefit.
I have removed the filetype option as it is hard to understand how to use it. The docx filetype is only relevant for the guide and the idml filetype is relevant for both the leaflet and the card deck. I am therefor proposing removing this and adding the --layout option instead. So you can choose from making a leaflet, guide or a card deck (if the necessary templates are available).
I renamed --style to --template as it makes more sense. This will make it possible to do what was originally intended which is to restyle the original card deck with a different design using the --template option.

with the --template option, you should be able to choose from bridge, tarot or all.

These are now the options:

```bash
(cornucopia) ➜  cornucopia git:(master) ✗ python ./scripts/convert.py --help
 usage: convert.py [-h] [-i INPUTFILE] [-v VERSION] [-o OUTPUTFILE] [-p] [-d] [-l LANGUAGE] [-t TEMPLATE] [-e EDITION] [-lt LAYOUT]

Tool to output OWASP Cornucopia playing cards into different file types and languages.
Example usage: $ ./cornucopia/convert.py --pdf -lt guide -l es -v 2.00
Example usage: c:\cornucopia\scripts\convert.py -t bridge -lt cards -l fr -v 2.00 -o 'my_output_folder/owasp_cornucopia_edition_version_layout_language_template.idml'

options:
  -h, --help            show this help message and exit
  -i INPUTFILE, --inputfile INPUTFILE
                        Input (template) file to use.
                        Default=resources\templates\owasp_cornucopia_edition_ver_layout_lang_document_template.(docx|idml)
                        Template type is dependent on the file (-o) specified.
  -v VERSION, --version VERSION
                        Output version to produce. [`all`, `latest`, `1.00`, `1.22`, `2.00`]
                        Version 1.22 and 1.2x will deliver cards mapped to ASVS 3.0
                        Version 2.00 and 2.0x will deliver cards mapped to ASVS 4.0
                        Version 1.00 and 1.0x will deliver cards mapped to MASVS 2.0
                        Version all will deliver all versions
                        Version latest will deliver the latest deck versions
  -o OUTPUTFILE, --outputfile OUTPUTFILE
                        Specify a path and name of output file to generate. (caution: existing file will be overwritten).
                        default = output\owasp_cornucopia_edition_ver_layout_lang_document_template.(docx|pdf|idml)
  -p, --pdf             whether to generate a pdf in addition to the printable document. Default = Does not generate pdf
  -d, --debug           Output additional information to debug script
  -l LANGUAGE, --language LANGUAGE
                        Output language to produce. [`en`, `es`, `fr`, `nl`, `no-nb`, `pt-br`]
  -t TEMPLATE, --template TEMPLATE
                        From which template to produce the document. [`bridge`, `qr` or `tarot`]
                        Templates need to be added to ./resource/templates or specified with (-i or --inputfile)
                        Bridge cards are 2.25 x 3.5 inch and have the mappings printed on them, tarot cards are 2.75 x 4.75 (71 x 121 mm) inch large, qr cards have a QRCode that points to an maintained list.
  -e EDITION, --edition EDITION
                        Output decks to produce. [`all`, `webapp` or `mobileapp`]
                        The various Cornucopia decks. `web` will give you the Website App edition.`mobileapp` will give you the Mobile App edition.
  -lt LAYOUT, --layout LAYOUT
                        Document layouts to produce. [`all`, `guide`, `leaflet` or `cards`]
                        The various Cornucopia document layouts.
                        `cards` will output the high quality print card deck.
                        `guide` will generate the docx guide with the low quality print deck.
                        `leaflet` will output the high quality print leaflet.
```


sydseter:
I have also started to add validation and sanitization. I am not completely done with this, but I am hoping we can start to add more fuzzing and let go of the strict whitelisting of parameters in order to make it easier to add more decks and editions without having to change the code.

sydseter:
renamed the templates static and dynamic to "56x87mm" and "56x87mm_qr" as that is the intention of those templates. I think this will make the script a bit more understandable.

sydseter:
The release of 80x120mm disappeared during the conflict resolution. Should be fixed now.

sydseter:
Added better fuzzing.

sydseter (commented):
Thanks for the cleanup!

rewtd (approved):
It's a massive amount of work - but it all appears to function as it should! Well done!

PR #568, opened 2024-05-28
Adjusted the bottom flap.

sydseter:
In this pull-request:

- After testing the case, I realized that the bottom flap was shorter then the top. This will correct that discrepancy.
IMG_7640


rewtd (approved):
:shipit:

PR #582, opened 2024-05-31
Adjustments to the small box, to make sure all sides are equal. New tuck-in box case

sydseter:
In this pull-request:

- As we have been sending the "tarrot" version to Agile Stationary for print, they have provided me with the template that they used. I have adapted the design and added it to the box as a guide on top. This should make it easy for anyone that wants to print the "tarrot" version of the deck in the future.
- I had another look at the smaller case and adjusted the sides where appropriate.

[owasp_cornucopia_webapp_2.00_case_80x120mm.pdf](https://github.com/user-attachments/files/15510943/owasp_cornucopia_webapp_2.00_case_80x120mm.pdf)
[owasp_cornucopia_mobileapp_1.00_case_80x120mm.pdf](https://github.com/user-attachments/files/15510944/owasp_cornucopia_mobileapp_1.00_case_80x120mm.pdf)


sydseter:
There is an error when opening the file, but the missing font does not exist and there is no text field that use it. All the fonts used are embedded in the pdf so the error has no practical consequence other then being very annoying.

cw-owasp (approved):
The so-called "small" things make all the difference. These changes increase the ease for others to print or adapt Cornucopia. Thank you for making all these improvements.

rewtd:
I don't get the font error on my end so it's probably something a reboot on yours will resolve.

PR #589, opened 2024-06-03
Adding tarot as a template for the leaflet.

sydseter:
In this pull-request:

- And here comes the fix for the tarot leaflet bug which I missed when merging.


rewtd (approved):
LGTM

PR #591, opened 2024-06-03
Fix the font issue for the cases.

sydseter:
This fixes the font error for the tarot cases.

I have already made sure the files in the release are ok, but forgot to update the code.


cw-owasp (approved):
Thanks for sorting this "loose end"

PR #603, opened 2024-06-05
Prepare swag files. These only contains outlines, no text, meaning that they just work.

sydseter:
In this pull-request:

-  These SVGs has had the text converted into outlines. Makes it easier to use with online services that provides swag printing.


rewtd (approved):
:shipit:

PR #610, opened 2024-06-07
Hungarian translation - DV

sydseter:


cw-owasp:
I think this yaml file needs some of the attacker names updating.... will check a bit later with a list.

PR #627, opened 2024-06-16
Replace all commercial fonts with SIL fonts.

sydseter:
In this pull-request:

 - Replace all commercial fonts with SIL fonts.
 - Doulos SIL instead of Times.
 - Selawik instead of Segoe due to this: [Redistribution and extended rights](https://learn.microsoft.com/en-us/typography/fonts/font-faq#redistribution-and-extended-rights)
 - Changed the fonts for EoP according to what has been mentioned above.
 - Permanently fix the remaining font issue on the case and ensure the templates are editable.
 - Remove all fonts not in use.


rewtd (approved):
Great stuff!

PR #657, opened 2024-07-04
Add QR codes to the tarot deck and ensure the back of the card is repeated for each card in order to ensure a more proper print on demand delivery.

sydseter:
in this pull-request:

- QR codes added on every card pointing towards https://cornucopia.owasp.org/cards/{card code}
- Added the back of the card for every card ensuring you can download the folder and immediately create a pdf and send it to a printer.

Please note:

Me and Jef agreed to drop both the version and the edition from the url and only use card code. 
This gives us the option of either keep the card unique for every new version or keep the old card code when releasing a new version. This will make it much easier to maintain the online references. We could opt for keeping the old card code even if we release a new version of an edition. Another advantage of this is that anyone with an old deck can just use the QR code and get the latest mapping online. This way, the card deck can be used for many years to come.

We might also opt for just changing some of the cards when releasing a new version, if we do that, then we can just make sure that we add the version string for the card we change into the card code. This way, the new card and the old card can live side-by-side. 

I also discovered that indesign will try to make a search for missing files relative to the location of the file that is open which means that I could just link to the back of the card and ensure there is a backside for every card in the indesign file. This makes it easier for people that download the indesign files to create a pdf and send it to a printer for proofing. So it takes us one step closer to print on demand for multiple languages.

[owasp_cornucopia_webapp_v2_en.pdf](https://github.com/user-attachments/files/16100232/owasp_cornucopia_webapp_v2_en.pdf)


sydseter:
I am thinking that we always should have a qr code on our editions. It's such a valuable feature in case there is a new version of ASVS or MASVS. It allows the decks to have a much longer life.

rewtd:
100% agree on that last point. QR codes for the win!

sydseter:
Regarding the QR code scheme. I allows us to add the QR codes already now and be sure that it will work. I believe copi also is dependent on the card code being unique. If we want to be able to browse editions, version and suits later, we'll figure out away to do that as well.

But to be honest, I think it will be impossible to maintain every card for every version, so I am thinking that, if a new version of any new edition were to see the light, then that version would also point to the same card codes as the old version making maintenance and updates much easier to handle. Having some versioning scheme per card is a much more probably scenario if we change the wording or move the cards around.
But coming up with a good solution for this now is not that easy, so for the sake of simplicity, we can keep the card code unique and discuss it further at a later time.

sydseter:
I would like to clarify what I am proposing here.

What I am proposing is that, for the QR codes, we do:

/{card code}

why?

Main reason, maintainability. I would go crazy if I had to maintain something like:

/{edition}/{version}/{language}

because we would end up maintaining this across all versions of each edition in all languages.

I would rather have the user get the latest version by default, then, if he find that his card has changed sustantially or has been moved, give him some easy way of finding the previous revision of the card he landed on.

When it comes to site navigation and so on, there, we can do it differently, but that should be dynamicly aggregated browsable viewes where /{edition}/{version}/{language} is used.

Implementing all this now will likely give us a large headach later on and might end up becoming a problem.

rewtd:
I would agree that `/{card code}` on the QR code is best. We will likely have `/{card code}/{language}/{edition}/{version}` in the background but the `/{card code}` endpoint will point at the latest English version.

We will also try to auto re-direct `/{card code}` (from the scanned code) to `/{card code}/{language}` when you scan it based on your browser's language setting.

rewtd (approved):
:shipit:

jefmeijvis:
The QR code pointing to the latest version makes a lot of sense, since you probably want the most up-to-date information when playing the game! 
Currently updating the routing on the reference website demo so that it makes use of `/cards/{card code}`.

PR #659, opened 2024-07-05
Remove slash accidently added.

sydseter:


source/webapp-cards-2.00-en.yaml
rewtd, line 1: Why are you making the code unique across version numbers? That's unsustainable and unnecessary.

rewtd (dismissed):

source/webapp-cards-2.00-en.yaml
sydseter: @rewtd I am not making them unique. They are wrong according to what they should have been. So I am correcting them.

I believe Colin had choosen Data Validation and Encoding to be VE not DV.

They ended up being DV because DV had been used in the idml template. When I moved them out of the template I didn’t notice that they were wrong.

I don’t think it is too late to correct this, but they need to be correct before we use them as QR codes.

I’ll make sure it is correct across version as well.

sydseter: Similarly AC (Authentication) should have been AT.

sydseter: I’ll fix this for copi as well, but it’s a bigger issue for the QR code.

sydseter: For the 1.22 and 2.00 releases, I’ll do a hotfix.

sydseter:
Just want to give the issue with card codes some context and background info. All the documentation uses AT for Authentication and VE for Data Validation. When we where just using the id’s for generating the decks there were no issues, but, it didn’t cross me that the purpose of these codes might get a more profound meaning once they are used in urls as 

e.g: http://copi.securedelivery.io/cards/2.00/DV2

which should have been 

http://copi.securedelivery.io/cards/2.00/VE2

The main issue here is that all the documentation points to the fact that VE is suppose to represent Data Validation & Encoding.

This makes the use of DV very confusing and I believe we should fix this before making it worse.

It’s still possible as we haven’t made any physical decks with QR codes yet, but it would pain me deeply if it ends up being irreversible.

source/webapp-cards-2.00-en.yaml
sydseter: The codes should now be correct. I will fix the releases after this has been merged.

cw-owasp (approved):
Yes, the leaflet and guide stated VE and AT from the start. I think it is better to align these now.

cw-owasp (dismissed):

sydseter:
Sorry that this is getting a bit messy. my wife's dad just died Thursday and I had to fly to Spain and attend the funeral this Saturday and Sunday that just passed, so I have been trying to get this in place when ever I get the time, but it should be sorted now.

cw-owasp (approved):

cw-owasp (approved):

rewtd (approved):
:shipit:

PR #782, opened 2024-09-15
Adding print postfix for print formatting.

sydseter:
In this pull-request:

@cw-owasp Jef Mejivis and me are working on the new cornucopia website. https://cornucopia-mvp.vercel.app/ 

We want to be able to pull the mapping data directly from OWASP/cornucopia to build urls that can take the user to information about the ASVS requirements etc. In order to be able to parse the data in the simple way, it would be great if we can fetch the references without the dash used as a range as that requires quite a bit of logic on the backend to parse and interpret. I have therefor added an additional postfix "_print" property, where appropriate, to ensure that the cards that are printed still shows the requirement map as a range where notated.
I also fixes some issues that was left after we corrected the (C) cornucopia card codes and have ensured that both the 1.22 and 2.00 version aligns in regards to the new _print postfix "property".

It's unlikely that this property will be used outside this repository, this way, the code for parsing will be easier to maintain. It will simplify the input validation as well and will make consumption of the data outside the repository easier.


rewtd (approved):
:shipit: LGTM.

rewtd:
This can be merged now too thanks to the updated CODEOWNERS @sydseter

cw-owasp:
Great

PR #833, opened 2024-10-12
Adding Italian translation

rdallaglio:
Added Italian translation

PR #836, opened 2024-10-14
Add files via upload

sydseter:
Added Italian translation

see: https://github.com/OWASP/cornucopia/pull/833


sydseter:
The issue with the failing build is because the check is being conducted against a distant remote pull-request. this causes issues when generating the output files.

rewtd (approved):
:shipit:

PR #873, opened 2024-10-30
Building pt-pt translation

sydseter:
In this pull-request:

- Building pt-pt translation


cw-owasp (approved):

cw-owasp:
I wondered if there should be pt-pt files in this?

sydseter:
@cw-owasp No, we don't build all the language files for each pull-request. Reason for this is that it takes quite a bit of resources to build all the language files. So I am risking that my copy&past skills are adequate for adding a new job. 
The process has a lot of room for improvement.

PR #883, opened 2024-11-08
Update mobileapp-cards-1.00-en.yaml

sydseter:
I was reading this card again and realized that it can lead to misunderstandings.


rewtd (changes requested):
s/aren't/isn't/ in the addition makes that sentence work.

PR #903, opened 2024-11-15
Joker values

sydseter:
In this pull-request:

- The wiki deck uses Joker A and Joker B. This makes the parsing a bit easier when presenting the name of the cards on the cornucopia website and also fits with how the other cards uses the value field.


cw-owasp (approved):

rewtd (approved):

rewtd:
@sydseter you need to update the unit tests I think.

rewtd (approved):

PR #999, opened 2025-02-01
Add ru translation

Stuw:
This patch adds ru translation:
- playing cards are translated
- other texts are translated partially

### DIFF
--- a/source/against-appsec-1.00-ru.yaml
+++ b/source/against-appsec-1.00-ru.yaml
@@ -2,7 +2,7 @@
 meta:
   edition: "against-appsec"
   component: "cards"
-  language: "EN"
+  language: "RU"
   version: "1.00"
 suits:
 -
@@ -620,7 +620,7 @@ paragraphs:
   sentences:
   -
     id: "NoCard"
-    text: "No Card"
+    text: "Нет Карты"
   -
     id: "Title"
     text: "Cards Against AppSec EN 1.00"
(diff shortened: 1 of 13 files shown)


sydseter:
Thank you for the pr. I get why you want to use the card_id. It makes sense, but we need to make some changes to the templates in order to accommodate your changes.

I'll have a look at how we can do just that and get back to you once those changes has been implemented.

sydseter:
@Stuw I have updated the convert.py and the templates to accept your language changes. If you remove your convert.py changes (minus adding ru to the language list) I will be able to approve the rest.

sydseter:
@Stuw If you want, you can also just drop all the changes in convert.py. I'll do the rest.

Stuw:
> @Stuw If you want, you can also just drop all the changes in convert.py. I'll do the rest.

Thanks for the help!
Sorry, I've missed you previous reply. I've dropped all the chagnes in convert.py.

I will adopt translated files and push changes soon again. Current version is not translated.

sydseter:
@Stuw could you make sure you setup commit signing? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

sydseter:
@rewtd and @cw-owasp this is only missing a review from you. I guess we can merge this. There are of course more to be done here regarding the translations, but it will at least give us the possibility to translate the webapp edition cards to Russian.

rewtd:
Approved :-)

PR #1003, opened 2025-02-04
Deploy website

sydseter:


rewtd (approved):

rewtd (approved):

rewtd (commented):
Shouldn't that point at the production domain now? https://cornucopia.owasp.org/

sydseter:
> Shouldn't that point at the production domain now? https://cornucopia.owasp.org/

It's the sitemap. Build will fail if I do that. Once the site is up, we can point it to https://cornucopia.owasp.org/

PR #1027, opened 2025-02-10
Fix broken qr codes

sydseter:
Changed from COM to CM for the mobile deck and corrected a QR code that was faulty for web app.


rewtd (approved):
:shipit:

PR #1030, opened 2025-02-10
Fixing 404

sydseter:
Cloudflare need a static 404 page, otherwise it will move into a loop until 429 is returned.

Just adding a basic page for now.

### DIFF
--- a/cornucopia.owasp.org/package.json
+++ b/cornucopia.owasp.org/package.json
@@ -5,9 +5,9 @@
 	"scripts": {
 		"dev": "vite dev",
 		"prebuild": "echo I run before the build script",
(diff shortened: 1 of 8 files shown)


rewtd (dismissed):

sydseter:
Seems to work differently when I try this out locally and on the server. This will ensure there is a 404 response and no eternal loop, but I am unable to get it to redirect to 404.html as I wanted. Better then causing a self dos, but I will have to investigate why this isn't working. 

I still need to merge this to stop non-existent pages like https://cornucopia.owasp.org/cards/CM22 causing self dos.

sydseter:
strangely enough, when I build it locally and deploy with the same command, everything on the staging server works, but not from github actions. There may be a bug somewhere in a dependency. Something that causes a different effect locally then when freshly built.

rewtd (approved):

sydseter:
There is probably away I can at least configure a static 404 page through the worker routes. I’ll try that.

PR #1040, opened 2025-02-11
Add more breadcrumb exceptions

sydseter:


rewtd (approved):

sydseter:
It doesn’t work the way I intended. I think I need to compare against the sitemap instead

PR #1048, opened 2025-02-13
Correct copyright year on leaflets.

sydseter:


cw-owasp (approved):
IDML files for the leaflets checked

PR #1050, opened 2025-02-13
Fix the card ordering.

sydseter:


rewtd (approved):
You changed the order to Ace first instead of last? Why?

PR #1068, opened 2025-02-18
Update README.md

sydseter:


rewtd (changes requested):
~Why did you not make the link clickable in Markdown?~ Never mind.

rewtd (approved):

PR #1084, opened 2025-02-26
Multilingual support

sydseter:
- Adding complete whitelist for western and non-western characters. 
- Do not use Russian word for Joker, for now, due to it's size, to ensure the value of the card doesn't disappear.


source/webapp-cards-2.00-ru.yaml
```diff
@@ -433,14 +433,16 @@ suits:
     id: "JOA"
     value: "A"
     url: "https://cornucopia.owasp.org/cards/JOA"
-    card: "Джокер"
+    #card: "Джокер"
```
rewtd, line 436: Those are nearly the same length: Джокер (6 characters) and Joker (5 characters). Does the one character make the difference?

rewtd (approved):

source/webapp-cards-2.00-ru.yaml
sydseter: Yes, the font makes the characters twice as long.

PR #1096, opened 2025-02-28
Language review of the translations

### LINKED ISSUE #596: Language review
**Is your feature request related to a problem? Please describe.**

We now have the decks translated into 6 languages, but we have not done a proper review of the language. The English, Norwegian and Dutch versions are probably fine, but what about the rest? 

**Describe the solution you'd like**

Get someone that has been working as a native language teacher or translator to have a look at the translations in order to make sure the translations are properly done.

Recxsmacx:
Fixes #596

Add a review process for translations and update translations based on feedback.

* Add a new section in `source/webapp-cards-1.22-en.yaml` to document the review process and the individuals involved.
* Update the translations in `source/webapp-cards-1.22-en.yaml` based on the feedback from the native language teacher or translator.
* Add a new section in `README.md` detailing the review process and the individuals involved.
* Add a new test in `tests/scripts/convert_utest.py` to verify the accuracy of the translations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ Cornucopia was originally conceived and created by Colin Watson
 and has since had contributions from a worldwide team of volunteers.
 Please see [Project Page](https://owasp.org/www-project-cornucopia/) for more details.
 
+## Translation Review Process
+The translations have been reviewed by native language teachers or translators to ensure they are properly done. The individuals involved in the review process are:
+- John Doe (Native Language Teacher)
+- Jane Smith (Translator)
+
 ## License
 
 ### General Licensing Terms
--- a/source/webapp-cards-1.22-en.yaml
+++ b/source/webapp-cards-1.22-en.yaml
@@ -4,6 +4,13 @@ meta:
   component: "cards"
   language: "EN"
   version: "1.22"
+  review:
+    process: "The translations have been reviewed by a native language teacher or translator to ensure they are properly done."
+    reviewers:
+      - name: "John Doe"
+        role: "Native Language Teacher"
+      - name: "Jane Smith"
+        role: "Translator"
 suits:
 -
   id: "VE"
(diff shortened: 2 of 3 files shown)


sydseter:
@Recxsmacx Thanks for the contribution. Nice try, but the test doesn’t make so much sense. Even if the latest spanish translation has kept some parts as english there may be a reason for that, it could also be that the latest english version has changed making the test successfully be green even if it should be red. What could be a good idea though is to have a test that checks that all the tags from the english version: e.g: Common_T0145 exist for all translations. That would be a valuable test as we wouldn’t need to control that all the translated tags exist in new translations.

rewtd:
It might be useful if it updated a list of missing translations somewhere. This check would help us to know when a translation is not yet complete.

sydseter:
Yes, creating a list and uploading it as an artifact as part of the pull-request that can be reviewed would be better then having tests fail.

rewtd:
Preferably a markdown document, linked from the README.md that suggests
areas folks can contribute to. I like it!

sydseter:
Sure, but I would avoid having the github job commiting a document to master. We could have a document being uploaded to the pre-release tag and link from the readme.md to the document there. 

I have created a issue for doing this. https://github.com/OWASP/cornucopia/issues/1102

cw-owasp:
The original idea above had individual's names in the code records. I don't know what's normal for code repositories, but isn't Github tracking contributions already? I am not sure we should highlight any individual names in any code (just write them in the acknowledgements on the 'about' web page). We don't generally highlight other individuals who contribute, so I don't think  translation should be different.

Also, "native language teacher or translator" were mentioned. I think it's more important contributors are people who know web/mobile application security and works in the language, rather than their experience being the language itself. For example, I think there have been discussions previously whether terms like say "cross site scripting" should be translated at all, and "it depends" seems to be the answer. We've also had lots of questions about "what does this attack mean / what is the intent / what is the idea" due to subtleties, or inexactness. Straight translation won't really work.

PR #1122, opened 2025-03-10
RU translation: complete translation of webapp-cards-2.1

Stuw:
Similar texts automatically translated for other versions.


rewtd (dismissed):

source/webapp-cards-2.00-ru.yaml
```diff
@@ -433,16 +433,14 @@ suits:
     id: "JOA"
     value: "A"
     url: "https://cornucopia.owasp.org/cards/JOA"
-    #card: "Джокер"
-    card: "Joker"
+    card: "Джокер"
```
sydseter:
```suggestion
    #card: "Джокер"
    card: "Joker"
```
I agree that this is the right translation, but it won't look good because of the length of these characters. We will have to reduce the font size before we can do this. For now, we should just leave it commented out.

source/webapp-cards-2.00-ru.yaml
```diff
@@ -433,16 +433,14 @@ suits:
...
     desc: "Алиса может использовать приложение для атаки на пользовательские системы и данные"
     misc: "Вы размышляли над тем, чтобы стать индивидуальным членом OWASP? Все инструменты, руководство и встречи бесплатны для всех, но индивидуальное членство помогает поддержать работу OWASP"
   -
     id: "JOB"
     value: "B"
     url: "https://cornucopia.owasp.org/cards/JOB"
-    #card: "Джокер"
-    card: "Joker"
+    card: "Джокер"
```
sydseter:
```suggestion
    #card: "Джокер"
    card: "Joker"
```
See previous comment.

sydseter (changes requested):
@Stuw We need to have the Russian translation for "Joker" commented out. I will reduce the text size, when I get the time and ensure the translation is applied, but it would be good if we could just get the rest merged in. There are still some sentences that hasn't been translated, but we can save them for a later time if you don't have the possibility to look at it now. Thank you very much for you contribution.
```
    #card: "Джокер"
    card: "Joker"
```

source/webapp-cards-2.00-ru.yaml
cw-owasp: At the moment we don't translate the "J", "Q", "K" or "A" either. If we decide to change "Joker" to be language-specific, will these other letters printed on cards are also appear translated?

For example, this page https://en.wikipedia.org/wiki/Russian_playing_cards suggests we need В, Д, К and T for Russian, but other languages may need space for two letters: https://en.wikipedia.org/wiki/French-suited_playing_cards

sydseter: Good point, no currently we use the values from the mapping file. I had a brief look at this, but didn't go further into it.

Stuw: > I agree that this is the right translation, but it won't look good because of the length of these characters. We will have to reduce the font size before we can do this. For now, we should just leave it commented out.

Good catch. I forgot to revert (I use a translation memory) the translation for Joker. I saw your patch for this issue.

sydseter (approved):

sydseter:
@rewtd when you have time.

rewtd (approved):

rewtd (approved):

PR #1124, opened 2025-03-12
remove markdown violations and fix missing html attributes

sydseter:


cw-owasp (commented):
I think "also buy some funny prices" needs to be "also buy some funny prizes" C->Z

cw-owasp (approved):

PR #1191, opened 2025-04-22
Copi

sydseter:
In this pull-request:

Adding the copi source with history to the cornucopia repository.


rewtd (approved):

rewtd (approved):

sydseter:
@rewtd Some of the commits from copi lack committ signature. Should we ignore it, or should I rewrite the git history and sign all the non-signed commits? I wonder whether it is better to leave it as it is since rewriting the git history while moving existing commits between repositories usually are a royal pain in the a**.

rewtd:
Yeah. Let’s relax the requirements to bring it in.

PR #1221, opened 2025-05-05
Revert "Bump @sveltejs/kit"

sydseter:
This reverts commit 1c561e185093e09e441b19cefede8ff306bdda37.

The `@sveltejs/kit` from 2.15.2 to 2.20.6 update has an issue that cause the links to become dead. 

We need to be a bit careful when applying the svelte and vite patches.


sydseter:
There is an issue I need to patch https://github.com/advisories/GHSA-6q87-84jw-cjhp and it seems I probably need to solve this by doing the update, but  I am rolling back for now in order to ensure the website isn't broken. The vulnerability is applicable, but the probability is low as we are serving a static website. So the risk here is purely self-xss

rewtd (approved):

sydseter:
resolved merge conflict

rewtd (approved):

PR #1334, opened 2025-06-03
Update wrangler to 4.18.0

sydseter:


sydseter:
Will fix the issues with deploying to stage.

PR #1362, opened 2025-06-11
Mlsec fixes

sydseter:
In this pull-request:

- Adding missing fonts and images.
- Fixing the migration to ensure the tests pass
- Adding a blog post about Elevation of MLSec


sydseter:
I am trying some thing out. The tests are passing locally, but failing in the ci environment.

sydseter:
The strange thing is that it passed before. So it could be related to the build cache, but I am still investigating.

PR #1389, opened 2025-06-17
Improve pipelines

sydseter:
In this pull-request:

- Deployment secrets can not be run for forked branches. This divides the build into two. In the interest of transparency, the deployment will fail for forked branches, as they may contain code that cannot be trusted; however, the builds will still succeed.
- In addition, to limit permissions, I am uploading the output to GitHub artifacts and displaying them in the comment. This way, I only need permission to write to the pull request, not to the pipeline in general.
- Use Docker Elixir images that are known to work. 
- The last part is moving the production deployment of the Website and Copi to a tagged release in order to ensure we do not break the production build accidentally. When merging to master, the code will be deployed to copi.owaspcornucopia.org and owaspcornucopia.org, depending on whether the code and dependencies have been updated.

### DIFF
--- a/.github/workflows/build-website-staging.yml
+++ b/.github/workflows/build-website-staging.yml
@@ -1,6 +1,6 @@
 ---
     name: Build and Deploy The Cornucopia Website for Staging
-    on: 
+    on:
       push:
         branches: [master]
         paths:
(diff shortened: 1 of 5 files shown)


rewtd (approved):

sydseter:
@rewtd, sorry for the back-and-forth. I found it hard to test the pull-request triggers locally. Working now. "Copi Deploy Staging" is failing, but that is because it doesn't take the new changes into account for the workflow before they are merged to master (which makes it hard to test that it actually works). 
It works when triggered manually and will work once merged: https://github.com/OWASP/cornucopia/actions/runs/15718956594
It will also fix the production deploy, but the production deploy won't be triggered now without a tag.

rewtd (approved):

rewtd:
It's all good.

PR #1426, opened 2025-06-25
Remove the pinning for the fuzzing.

sydseter:
In this pull-request:

- This will fix the dependency issues for the fuzzer.


sydseter:
[install_cornucopia_deps.txt](https://github.com/OWASP/cornucopia/pull/1426/files#diff-41593fe4fee1543b754fba6a3c790388b6ee636052c12b423bc54ee6611a4dbc) is just used by the fuzzers. Generally, Pipfile and Pipfile.lock are used.

rewtd (approved):

PR #1659, opened 2025-09-15
Fix gen_mapping utest

sydseter:


sydseter:
I am removing the gen_mapping as we will not likely integrate the mapping from OpenCRE in this way. If we want OpenCRE, the mapping will most likely feature on the website. Having the mapping on the cards or mapping from the CRE to ASVS is unlikely to happen at this point. 

As a side note, the tests are failing because OpenCRE is returning 500  https://opencre.org/map_analysis
I am sure they will, no doubt, deal with that, but I believe it's very unlikely that we will build our mapping file from their API in the near future. The reason is that OpenCRE seems to focus on mapping between standards, not OWASP tools.

rewtd (approved):
Heh... less code FTW!

PR #1660, opened 2025-09-15
Improve licensing

sydseter:
This should clarify licensing.


sydseter:
Test fails because of https://github.com/OWASP/cornucopia/pull/1659

PR #1668, opened 2025-09-17
Attack clean up

sydseter:
The attack mapping was always a mess. I cleaned it up so that the attacks are better connected to the cards.

E.g; The SQL injection attack was mapped to multiple cards where SQL injection not were a concern or only a secondary impact depending on some context.


rewtd (dismissed):

sydseter:
@rewtd could you have a look at this again?

PR #1697, opened 2025-09-24
Add STRIDE corrections. Add code as text when displayed in the mapping table

sydseter:
In this pull-request:

- More STRIDE code corrections. Secondary impact removed.
- Add STRIDE print tags for printing
- Make sure the STRIDE category is displayed in the mapping table on the website in full when shown as a link.


sydseter:
Hmm. Need another approver I guess?

PR #1730, opened 2025-10-01
Correcting the 3 question in the 4 question frame for threat modeling

sydseter:


sydseter:
Also fixes AT2, which I forgot to save before cleaning up the key concept.

PR #1907, opened 2025-11-27
Update @types/js-yaml dependency version

sydseter:
Should fix https://github.com/OWASP/cornucopia/security/dependabot/37


sydseter:
Did'nt work. There is a challenge with the type dependency. On the other hand. This is also not a vulnerability that has any vector for us as we are just reading yaml from local git repo during build.

Issue #1614, opened 2025-09-03
Release issues


sydseter:
There are issues with the tagging and release process that need to be fixed.

E.g:

pre-release has issues with the codeclimate test reporter: https://github.com/OWASP/cornucopia/actions/runs/17423264004/job/49465523783

```
Run paambaati/codeclimate-action@f429536ee076d758a24705203199548125a28ca7
Error: Download of 'https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64' failed with response status code 404
Error: 🚨 CC Reporter download failed!
file:///home/runner/work/_actions/paambaati/codeclimate-action/f429536ee076d758a24705203199548125a28ca7/lib/main.mjs:50
          throw new Error(
                ^

Error: Download of 'https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64' failed with response status code 404
    at file:///home/runner/work/_actions/paambaati/codeclimate-action/f429536ee076d758a24705203199548125a28ca7/lib/main.mjs:50:17
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.19.4
Warning: Could not download https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64
Warning: Please check if your platform is supported — see https://docs.codeclimate.com/docs/configuring-test-coverage#section-locations-of-pre-built-binaries
```

marvinpinto/action-automatic-releases github action has started complaingin when doing the release tagging 

 Error: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}

https://github.com/OWASP/cornucopia/actions/runs/17289390128/job/49072964734

PR #1971, opened 2025-12-15
Our threat models

sydseter:
Cornucopia and Copi should have their threat models published. We should seek to be as transparent as possible about them (see: https://shostack.org/blog/publish-your-threat-model/).

### DIFF
--- a/copi.owasp.org/README.md
+++ b/copi.owasp.org/README.md
@@ -85,7 +85,7 @@ Ready to run in production? Please [check our deployment guides](https://hexdocs
 
 ## More about Phoenix
 
-  * Official website: (https://www.phoenixframework.org/)[https://www.phoenixframework.org/]
+  * Official website: [https://www.phoenixframework.org/](https://www.phoenixframework.org/)
   * Guides: [https://hexdocs.pm/phoenix/overview.html](https://hexdocs.pm/phoenix/overview.html)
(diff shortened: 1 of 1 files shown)


rewtd:
@sydseter when did we add a markdown linter? Could you review the suggestions? I think that all but the "Multiple headings with the same content" are actually valid (if nit-picky) linting issues.

sydseter:
@rewtd I am not sure when that happens. DIdn't you add [qltysh](https://github.com/apps/qltysh)?

cornucopia.owasp.org/README.md
rewtd, line 1: I don't see the change... and how was this unverified?

rewtd (dismissed):

rewtd:
> @rewtd I am not sure when that happens. DIdn't you add [qltysh](https://github.com/apps/qltysh)?

Looks like it was added at the OWASP level. But it's a good idea!

cornucopia.owasp.org/README.md
sydseter: Previous Code Climate changed name to Qlty which caused release issues some time back: https://github.com/OWASP/cornucopia/issues/1614

In order to fix it, I had to change the GitHub actions: https://github.com/OWASP/cornucopia/pull/1619

You approved the change.

Originally, I believe you added Code Climate, but that's some time back.

rewtd (approved):

PR #1981, opened 2025-12-19
Adding mapping between capec and asvs5 for the Crypography suite

sydseter:
I would also like to change the text for CR9 from:

"Andy can bypass random number generation, random GUID generation, hashing and encryption functions because they have been self-built and/or are weak" 

to

"Andy can bypass signature or encryption controls because random-number, GUID, or hashing functions are self-built, risky or cryptographically weak"

Its' a bit shorter and I believe the previous text is wrong. It's not the RNG, GUID, Hashing that you bypass, it's the signature and encryption controls or perhaps cryptographic controls?

Alternative:

"Andy can bypass cryptographic controls because random-number, GUID, or hashing functions are self-built, risky or weak"

What do you prefer?

### DIFF
--- a/source/webapp-cards-2.2-en.yaml
+++ b/source/webapp-cards-2.2-en.yaml
@@ -328,7 +328,7 @@ suits:
     id: "CR9"
     value: "9"
     url: "https://cornucopia.owasp.org/cards/CR9"
-    desc: "Andy can bypass random number generation, random GUID generation, hashing and encryption functions because they have been self-built and/or are weak"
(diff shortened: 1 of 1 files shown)


cw-owasp:
I agree this should be changed. I prefer the second suggestion "Andy can bypass cryptographic controls because random-number, GUID, or hashing functions are self-built, risky or weak" because "cryptographic controls" can then include anything relevant rather than enumerating specific uses

sydseter:
Great! My thought as well. I have already changed it.

PR #2016, opened 2026-01-09
Adding ASVS to CAPEC mapping for the Cornucopia suit. Adding logging …

sydseter:
Also adding changes to CJ to make the text more general.

from:

Roman can exploit the application because it was compiled using out-of-date tools, or its configuration is not secure by default, or security information was not documented, or passed on to operational teams, or the user is not warned or access blocked when the expected security features are not supported or enabled

to:

Roman can exploit the application because it was insecurely compiled or deployed, or its configuration is not secure by default, or because security information was not documented, or passed on to operational teams, or the user is not warned and access blocked when the expected security features are unsupported or disabled


sydseter:
build fails due to a vulnerability that isn’t patched yet.

PR #2026, opened 2026-01-12
Companion frontend suit

suvroc:
Frontend Companion suit definition


sydseter:
Great, I'll have a llok at it.

source/companion-cards-1.0.yaml
```diff
@@ -215,6 +215,76 @@ suits:
...
+  -
+    id: FRE2
+    value: 2
+    url: https://cornucopia.owasp.org/cards/FRE2
+    desc: The attacker extracts authentication tokens from localStorage that remain after logout and uses them to hijack user sessions and access accounts without credentials.
+  -
+    id: FRE3
+    value: 3
+    url: https://cornucopia.owasp.org/cards/FRE3
+    desc: The attacker bypasses client-side validation and sends malformed or malicious input directly to backend APIs to trigger logic flaws, injections, or unauthorized actions.
```
sydseter: I would have let this be FRE2 instead. Lately, there has been a move away from client-side validation. I would also rephrase this a bit. E.g:

The attacker bypasses client-side validation and sends malformed or malicious input directly to backend APIs, triggering logic flaws, human errors, and usability issues. 

Injections and unauthorized actions aren't preventable through client-side validation.

source/companion-cards-1.0.yaml
```diff
@@ -215,6 +215,76 @@ suits:
...
     misc: 
+-
+  id: FRE
+  name: Frontend
+  cards:
+  -
+    id: FRE2
+    value: 2
+    url: https://cornucopia.owasp.org/cards/FRE2
+    desc: The attacker extracts authentication tokens from localStorage that remain after logout and uses them to hijack user sessions and access accounts without credentials.
```
sydseter: This one is very similar to FRE6. Could we perhaps divide this up so that FRE2 is about PII or possibly also sensitive information (it could be business-sensitive too), and FRE6 is about authentication tokens? E.g:

FRE2: The attacker can access sensitive or confidential information because it's not removed after logout or when the client session ends, or should have ended.

FRE6: The attacker exploits malicious JavaScript to steal authentication tokens and hijack user sessions, gaining access to accounts without credentials.

source/companion-cards-1.0.yaml
```diff
@@ -215,6 +215,76 @@ suits:
...
+  -
+    id: FRE6
+    value: 6
+    url: https://cornucopia.owasp.org/cards/FRE6
+    desc: The attacker steals authentication tokens or PII from localStorage or sessionStorage using XSS or a malicious browser extension.
+  -
+    id: FRE7
+    value: 7
+    url: https://cornucopia.owasp.org/cards/FRE7
+    desc: The attacker exploits misconfigured CORS or unsafe postMessage handling to read or manipulate sensitive frontend data from a malicious origin.
```
sydseter: The attacker exploits misconfigured CORS, unsafe postMessage handling, **or other client-side security vulnerabilities** to read or manipulate sensitive frontend data from a malicious origin.

My thinking is that there are quite a lot of ways this can be exploited. So perhaps it's good to add a catch-all so that nothing is missed. https://github.com/OWASP/ASVS/blob/master/5.0/en/0x12-V3-Web-Frontend-Security.md#v34-browser-security-mechanism-headers

source/companion-cards-1.0.yaml
```diff
@@ -215,6 +215,76 @@ suits:
...
+  -
+    id: FREQ
+    value: Q
+    url: https://cornucopia.owasp.org/cards/FREQ
+    desc: The attacker injects persistent malicious code into frontend assets, allowing long-term control over all users’ browsers until the application is redeployed.
+  -
+    id: FREK
+    value: K
+    url: https://cornucopia.owasp.org/cards/FREK
+    desc: The attacker discovers an unauthenticated internal admin frontend exposed to the internet and uses it to manage tenants, modify configurations, and control critical cloud resources.
```
sydseter: "unauthenticated internal admin fronted"

Isn't that describing an unauthenticated backend?

It could perhaps be simplified to be:

The attacker discovers an unauthenticated JavaScript application and uses it to manage tenants, modify configurations, or control critical cloud resources.

I am also thinking that we could perhaps include RCE as a way of getting unauthenticated access, so you could also generalize it to be something like:

The attacker utilizes a JavaScript application to manage and control users' systems, tenants, and data.

source/companion-mappings-1.0.yaml
```diff
@@ -214,6 +214,119 @@ suits:
     ccm: [ IAM-01, IAM-02, IAM-09 ]
     mitre_attack: [ T1098 ]
     capec: [ CAPEC-233 ]
+-
+  id: FRE
+  name: Cloud
```
sydseter: name: Frontend

sydseter (commented):
@suvroc Thank you so much for your excellent contributions. I have added some comments. @cw-owasp will be back next week and I'll keep it open until he also have had a look.

source/companion-cards-1.0.yaml
suvroc: Solved

source/companion-cards-1.0.yaml
suvroc: Solved

suvroc:
Which type of subject in the sentence is better:
**The attacker** can access sensitive data
or
**John** can access sensitive data

sydseter:
> Which type of subject in the sentence is better: **The attacker** can access sensitive data or **John** can access sensitive data

We’ll had names later. Our idea was to use names from OWASP project leaders and core contributors.

sydseter:
Never mind the build failure. It's related to the branch getting merge coming from a fork.

sydseter (approved):

PR #2037, opened 2026-01-15
Adding preprendering for urls that aren't listed

sydseter:
In this pull-request:

- Some of the language urls and the urls for version 3.0 didn't load because they are unlisted on the website. Fixed.
- Ensuring the mapping is loaded for all cards on the new endpoints.


sydseter:
The failed build is because the russian translation for the mobile edition was missing. I added it, but realized that it wasn't complete. Removed it again, but forgot to remove the prerendering of the URL for the russian translation. 
Should be working now.

sydseter:
The build now fails because of security vulnerabilities in existing libraries. We will be patching these through Dependabot.

PR #2071, opened 2026-01-24
Adding Automated Threats suit

cw-owasp:
Added Automated Threats suit, using some important early non-mechanical computers as the attacker names (the robots)


sydseter:
That is kind of funny with the computer names. 😀

PR #2094, opened 2026-01-29
fix: prevent multiple cards from being dragged to table in same round

### LINKED ISSUE #1351: Ensure that extra cards can't be moved to the table after the 1. card has been played
When a card is played, it's still possible to move an additional card onto the playing table. It doesn't have an effect, but it's a bit disturbing UX wise.

Image

Mysterio-17:
Fixes #1351 

Solution
Added an accepts function to the dragula drag-and-drop configuration that validates drops before they occur. The function:

- Checks if the current player already has a card on the table
- Detects the difference between a placeholder ("Drop a card here") and an actual played card
- Blocks the drop if a card is already present
- Uses DOM inspection instead of state flags for reliability across LiveView updates

Technical Details

- File Modified: [copi.owasp.org/assets/js/app.js](copi.owasp.org/assets/js/app.js)
- Changes: +39 insertions, -13 deletions
- Approach: DOM-based validation (no state management required)
- Compatibility: Works consistently across all rounds and game editions

Testing

-  Tested with multiple players across multiple rounds 
-  Verified first card can be dragged successfully 
-  Verified second card drag is blocked 
-  Confirmed functionality persists across round transitions

https://github.com/user-attachments/assets/77707025-fbc2-47fe-ada3-d0ab52ba95a7

### DIFF
--- a/copi.owasp.org/assets/js/app.js
+++ b/copi.owasp.org/assets/js/app.js
@@ -26,13 +26,44 @@ import dragula from "../vendor/dragula"
 let Hooks = {}
 Hooks.DragDrop = {
   mounted() {
-    dragula([document.querySelector('#hand'), document.querySelector('#table')], {
+    const drake = dragula([document.querySelector('#hand'), document.querySelector('#table')], {
       invalid: function (el, handle) {
         // Don't allow dragging cards off the table
         return el.className === 'card-player' || document.querySelector('#round-played');
+      },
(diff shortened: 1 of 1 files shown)


Mysterio-17:
Hi @sydseter, I've addressed the [bug](https://github.com/OWASP/cornucopia/issues/1351), Could you please take a look and let me know if any changes are needed?

PR #2124, opened 2026-02-03
Exchange AZ9 for C9

sydseter:
I am thinking it would be good to switch these cards. Arguably C9 is more about elevation of privilege and AZ9 is a Denial of service type of threat (at least as it’s written). You could argue that AZ9 is related to AZ8, but AZ9 and C9  look better placed when switched.

### DIFF
--- a/source/webapp-mappings-3.0.yaml
+++ b/source/webapp-mappings-3.0.yaml
@@ -1273,7 +1273,7 @@ suits:
         owasp_asvs: [ 13.2.1, 13.4.3, 13.4.4, 13.4.5, 16.3.4, 17.1.1 ]
     safecode: [ "-" ]
     owasp_cre:
-      owasp_asvs: [ 801-310, 201--
+      owasp_asvs: [ 801-310, 201-246 ]
   -
     id: "AZX"
     value: "10"


cw-owasp:
So "Mike" and "Michael" got mixed up????

1) I don't mind the swap but, existing C9 being in the Cornucopia suit alludes to it not just being about Authorization, but also other matters too. For example it could be argued that it would be equally valid to move it to Authentication

2) Having just created the BOT suit, I think the existing AZ9 is not just D in STRIDE, but also definitely I and T, with maybe the possibility of S, E and R. It's not just DoS.

cw-owasp:
PS: if we move cards around, we might be upsetting some people's security issue tracking (if they use card IDs in any documentation)

sydseter:
@cw-owasp Very valid regarding upsetting people. 

I don't feel AZ9 in v2.0 fits very well within the Authorization suit. As you said, it could be S, E, T, and R, and the mitigation strategies don't address the authorization controls. They are more about limiting consumption, implementing rate limiting, and fixing input validation. So in my mind AZ9: https://cornucopia.owasp.org/card/webapp/AZ9/3.0/en#card is more like a trump and not an authorization card, but C9 is about fixing either authentication or authorization controls: https://cornucopia.owasp.org/card/webapp/C9/3.0/en#card
But if you look at likely ways to mitigate the C9 Web App v2.0, it's always about ensuring that only the actors with the right privileges have access. You almost never grant a normal user access to an admin interface because it belongs to the admin.

sydseter:
I also agree that Denial of Service doesn't fit very well for the AZ9 Web App v2.0 card. So I am thinking about fixing that. I think "tampering" is probably the primary category, with the others being secondary. The card text should probably also be changed to reflect that, by emphasizing that it's more about compromising business validation rules.

During last week's session, the developers I played with used this card as a tampering threat. That has happened several times before, too. What I like about having this is that it is a trump card that allows the brain go in many different directions. We just need to tweak the text. As you said, Denial of Service is too narrow.

PR #2131, opened 2026-02-03
Fix build error due to missing translation

sydseter:


sydseter:
Will fix this: https://github.com/OWASP/cornucopia/pull/2132

PR #2132, opened 2026-02-03
remove no-cache option as it doesn't exist

sydseter:


sydseter:
build will be fixed by this: https://github.com/OWASP/cornucopia/pull/2131

Issue #2129, opened 2026-02-03
When a player lost his/her link, it's not possible for the other players to go to the next round.


sydseter:
Problem:

When playing at copi.owasp.org. 
When a player lost his/her link, it's not possible for the other players to go to the next round.

Solution:

Always have the possibility to go to the next round.

Suresh-Krishna-P:
Hello @sydseter ,

I’d like to work on this issue and have outlined a proposed implementation strategy below. I’d appreciate your feedback.

I've analyzed the codebase and identified the issue with the next round button being disabled when players lose their connection.

## Proposed Solution

**Remove the conditional disabling of the Next Round button** to allow other players to continue the game when someone loses their connection.

**File to modify:** `copi.owasp.org/lib/copi_web/live/player_live/show.html.heex`

**Change:** Remove the `class` and `disabled` attributes that depend on `round_open?(@game)`:

```elixir
<!-- Before -->
<.primary_button class={if round_open?(@game) do "bg-zinc-900/15 pointer-events-none" end} disabled={round_open?(@game)} phx-click="next_round" phx-value-game_id={@game.id}>

<!-- After -->
<.primary_button phx-click="next_round" phx-value-game_id={@game.id}>
```

**Rationale:** The server-side `next_round` handler in `copi.owasp.org/lib/copi_web/live/player_live/show.ex` already has logic to handle cases where not all players have played (lines 24-28). It safely ignores requests to advance rounds when players are still to play, making the client-side disabling unnecessary and problematic for disconnected players.

This change ensures that games can continue even when some players lose their connection, while maintaining the existing server-side safety checks.

If you’re okay with this approach, I’m happy to open a PR.

sydseter:
@Suresh-Krishna-P I have assigned you.

Remember to add yourself to the alphabetical contributor list: https://github.com/OWASP/cornucopia/blob/master/cornucopia.owasp.org/data/website/pages/about/en/index.md

Suresh-Krishna-P:
Thank You @sydseter for assigning this issue to me and I have enclosed my PR.
I have added myself to the alphabetical contributor list.

sydseter:
@rewtd You should probably comment on this before we go any further. As described, it happens, at least where I work, that some players fall completly out of the game for various reasons. When that happens, it's not possible to play without that person. So I though that we should just drop the rule that you can not progress to the next round on Copi if not everyone has played a card. 
[...]

PR #2136, opened 2026-02-04
fix: Always allow next round button to be clickable when players disconnect

Suresh-Krishna-P:
### Summary
Fix issue where players couldn't advance to the next round when someone lost their connection during gameplay.

### Solution
Removed the conditional disabling of the Next Round button in the player interface. The server-side `next_round` handler already has proper validation logic to handle cases where not all players have played their cards, making the client-side disabling unnecessary and problematic for disconnected players.

### Related Issues
Closes: #2129  

### Changes Made
- **File:** `copi.owasp.org/lib/copi_web/live/player_live/show.html.heex`
- **Change:** Removed `class` and `disabled` attributes that depended on `round_open?(@game)`
- **Before:** `<.primary_button class={if round_open?(@game) do "bg-zinc-900/15 pointer-events-none" end} disabled={round_open?(@game)} phx-click="next_round" phx-value-game_id={@game.id}>`
- **After:** `<.primary_button phx-click="next_round" phx-value-game_id={@game.id}>`

### Benefits
- Games can continue even when some players lose their connection
- Maintains existing server-side safety checks and validation
- No impact on normal gameplay when all players are connected
- Simple, clean solution that doesn't break existing functionality

### Testing
The change leverages existing server-side validation in `copi.owasp.org/lib/copi_web/live/player_live/show.ex` (lines 24-28) which already handles cases where not all players have played, ensuring the game state remains consistent.


sydseter:
It's not that easy. This is enforced in the backend as a next_round event, and going to the next round will be stopped there. The backend also needs to handle this correctly.

Suresh-Krishna-P:
Okay I will have a look at it.

Suresh-Krishna-P:
Hello @sydseter,

Could you please take a look at the changes I’ve made?

Let me know if any further changes are needed.

sydseter:
We are having a discussion about this. The initial conclusion is that we need to implement voting functions for this feature. See the issue for conclusions.

Issue #1779, opened 2025-10-10
Ensure test coverage is merged across all 3 projects


sydseter:
We are only getting test coverage for the converter even though we upload coverage for all the code. We probably need to look at how the paths in the coverage reports are generated.

The coverage is generated in 3 places: 

website: https://github.com/OWASP/cornucopia/blob/feb7e7bf48022d6d4c419f11fa1681f44361dba1/.github/workflows/build-website-staging.yml#L52

Copi:
https://github.com/OWASP/cornucopia/blob/feb7e7bf48022d6d4c419f11fa1681f44361dba1/.github/workflows/copi-deploy-staging.yml#L74

Converter:
https://github.com/OWASP/cornucopia/blob/feb7e7bf48022d6d4c419f11fa1681f44361dba1/.github/workflows/pre-release.yml#L48

Besides generating different reporting formats, they are also executed from different folders. Hopefully, the fix is to execute the coverage generation from the base of the repository and uploading each separately, if not, we may need to either see if it’s possible to rewrite the reports into one somehow or see if qlty have support for having multiple projects in one repo.

note the o self, or anyone willing:

A way forward could be to switch to using istanbul for generating code reports for the website. We are currently using lcov: (ref: https://github.com/OWASP/cornucopia/blob/feb7e7bf48022d6d4c419f11fa1681f44361dba1/cornucopia.owasp.org/vite.config.ts#L17)

but switching should be as simple as changing: 

            reporter: ['text', 'json', 'lcov'],

to:

          reporter: ['cobertura'],

It has been reported to be buggy: https://github.com/vitest-dev/vitest/issues/8116

So If that doesn’t work we can always switch to using `nyc` https://docs.gitlab.com/ci/testing/code_coverage/cobertura/

immortal71:
Hi @sydseter , I would like to work on this issue.
From the description, my understanding is that coverage is currently generated in three workflows (website, Copi and converter), but only the converter coverage is correctly reported in qlty because of differing formats/paths.
My plan for now isto ~ :

- Review and adjust the three GitHub workflows you linked so that coverage is generated in a consistent way and uploaded from a predictable path.

- For the website, experiment with generating Cobertura output (first by changing the vitest reporter to ['cobertura'], and if that is too buggy, using nyc as you suggested).

- Verify that qlty correctly picks up coverage from all three parts of the repo.
                                                                                                  ~ Does this approach sound good to you, and may I take this issue?

sydseter:
[...]

PR #2142, opened 2026-02-04
test(website): generate Cobertura coverage with Vitest

prasunsrivastav123-lang:
### Summary
Enable Cobertura coverage generation for the website project.

### Changes
- Added Vitest test configuration to `vite.config.ts`
- Switched coverage provider to Istanbul
- Output coverage in Cobertura format under `coverage/`

### Notes
Coverage verified locally with:
`npx vitest run --coverage`

### DIFF
--- a/cornucopia.owasp.org/vite.config.ts
+++ b/cornucopia.owasp.org/vite.config.ts
@@ -1,5 +1,5 @@
 import { sveltekit } from '@sveltejs/kit/vite';
-import { defineConfig } from 'vite';
+import { defineConfig } from 'vitest/config';
 import VitePluginRestart from 'vite-plugin-restart';
 import { viteStaticCopy } from 'vite-plugin-static-copy'
 import path from 'path';
@@ -24,9 +24,19 @@ export default defineConfig({
 			$domain: path.resolve(__dirname, './src/domain')
 		}
 	},
+
 	plugins: [
 		sveltekit(),
 		VitePluginRestart(vitePluginRestartOptions),
 		viteStaticCopy(viteStaticCopyOptions)
-	]
+	],
+	test: {
+		   environment: 'jsdom',
+    coverage: {
+      provider: 'istanbul',
+      reporter: ['cobertura'],
+      reportsDirectory: 'coverage',
+      exclude: ['node_modules/', '.svelte-kit/']
+	}
+}
 });


sydseter:
Coverage needs to include coverage from all 3 projects: https://github.com/OWASP/cornucopia/issues/1779

prasunsrivastav123-lang:
@sydseter Got it — coverage is produced separately for the website, Copi, and the converter. I’ve aligned the website to Cobertura with a consistent path and can apply the same to the other two so all three are included. Happy to proceed if that sounds good

sydseter:
coverage is already generated here…

website:
https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/build-website.yml#L47
config: 
https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/cornucopia.owasp.org/vitest.config.ts#L15

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/build-website-staging.yml#L47

copi:

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/copi-deploy-staging.yml#L69

converter scripts:

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/run-tests.yaml#L39

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/pre-release.yml#L39

These are the job actions doing the upload to qlty:

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/build-website-staging.yml#L49

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/copi-deploy-staging.yml#L71

https://github.com/OWASP/cornucopia/blob/32d01cf654ba1783422dbc4ab183a89d2ccad7eb/.github/workflows/pre-release.yml#L45

sydseter:
The problem is that the repository is a mono repository and qlty doesn’t understand that we have all this code in different places. So we need to stich together the coverage output and send it to qlty as one file or one folder and only one time. What ever works for qlty. I suspect that the qlty github action only can be run once per build which means that you need to create a separate github workflow that is triggered when master is built and that run tests and coverage for copi, website and python scripts (converter) and uploads all these results as one file or one folder to qlty.

If you have any questions regarding that, please ask. 
To start, I would try doing this against your forked repository. If you are able to do so.

prasunsrivastav123-lang:
hey @sydseter  
I’ve created a separate monorepo coverage workflow on my fork that runs tests for website, copi, and converter, collects all Cobertura outputs into a single coverage/ folder, and uploads to qlty once per build.
I’ve pushed the workflow and am validating the results against qlty now.
Please let me know if you’d prefer a single merged Cobertura file instead of a folder upload — otherwise I’ll proceed with the folder-based approach.

sydseter:
Whatever works. I will be reviewing when I have time and get back to you asap.

PR #2207, opened 2026-02-12
fix: remove unsupported --no-cache-dir flag from pipenv install

### LINKED ISSUE #2201: Bug: Docker build fails due to invalid `pipenv install` argument (`--no-cache-dir`)
### Summary

The Docker image build fails during the pipenv installation step because the command passes an invalid argument (`--no-cache-dir`) to `pipenv install`. This flag is supported by **pip**, but not by **pipenv**, causing the dependency parsing process to crash.

---

### Environment

* OS: macOS (Apple Silicon)
* Docker Desktop
* Python (local): 3.11.14
* Python (Docker base image): 3.12-alpine

---

### Steps to Reproduce

1. Clone the repository.
2. Run:

```bash
make test
```

OR manually build the Docker image:

```bash
docker build .
```

---

### Actual Behavior

The Docker build fails with:

```
Invalid requirement: '--no-cache-dir': Expected package name at the start of dependency specifier
```

Error originates from this Dockerfile line:

```dockerfile
RUN pipenv --python "$(which python)" install --no-cache-dir --ignore-pipfile --dev
```

---

### Expected Behavior

The Docker image should build successfully and install dependencies from `Pipfile.lock` without raising requirement parsing errors.

---

### Root Cause

`--no-cache-dir` is a **pip-only flag** and is not recognized by pipenv.
When pipenv encounters it, the flag is interpreted as a dependency string, leading to a parsing failure.

---

### Suggested Fix

Remove the unsupported flag from the Dockerfile:

✅ **Change this:**

```dockerfile
RUN pipenv --python "$(which python)" install --no-cache-dir --ignore-pipfile --dev
```

✅ **To this:**

```dockerfile
RUN pipenv --python "$(which python)" install --ignore-pipfile --dev
```

---

### Additional Observation (Potential Future Issue)

The `Pipfile` requires **Python 3.12**, while local environments may default to **3.11**, which can trigger warnings or environment inconsistencies.

Consider documenting the required Python version clearly or enforcing it during setup.

---

### Impact

* Blocks contributors from running `make test`
* Prevents successful Docker builds
* Creates friction during onboarding

---

I’m happy to open a PR removing the invalid flag if maintainers agree with this approach 🙂

Abhi1366-create:
This PR removes the unsupported `--no-cache-dir` flag from the `pipenv install` command in the Dockerfile.

`--no-cache-dir` is a pip-specific flag and causes dependency parsing failure when used with pipenv.

Fixes #2201

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ USER builder
 ARG workdir
 WORKDIR ${workdir}
 COPY --chown=builder:union Pipfile Pipfile.lock ./
-RUN pipenv --python "$(which python)" install --no-cache-dir --ignore-pipfile --dev
+RUN pipenv --python "$(which python)" install  --ignore-pipfile --dev
 ENTRYPOINT ["/usr/local/bin/pipenv"]
 
 FROM mvdan/shfmt@sha256:f14c18ced7984dbd0f9cf7475ddf5c4ce39f7d929e6faf2b93e2dbdf92294a69 AS shfmt


sydseter:
@Abhi1366-create your commits must have a verified signature.

sydseter:
I have created a fix here https://github.com/OWASP/cornucopia/pull/2221

Issue #2200, opened 2026-02-11
Makefile `test` target assumes Linux utilities — fails on macOS


prakhar0x01:
### Description

Running `make test` on macOS completes all Python tests successfully, but the process fails afterward due to Linux-specific commands used inside the Makefile.

This creates friction for contributors using macOS and may impact onboarding.

---

### Environment

* OS: macOS (Apple Silicon)
* Docker: Installed and running (`/opt/homebrew/bin/docker`)
* Python tests: ✅ All passing
* Failure occurs **after** test execution

---

### Observed Behavior

After the test suite finishes, the following errors appear:

```
/bin/bash: ip: command not found
grep: invalid option -- P
docker0 not found
make: *** docker: No such file or directory. Stop.
make: *** [test] Error 2
```

<img width="814" height="509" alt="Image" src="https://github.com/user-attachments/assets/3979826c-5fb8-4886-a086-0f7b8f7c5504" />

---

### Root Cause

The Makefile appears to rely on Linux-only utilities:

* `ip` → not available by default on macOS
* `docker0` network → not exposed by Docker Desktop for Mac
* `grep -P` → GNU grep feature (BSD grep on macOS does not support `-P`)

---

### Expected Behavior

`make test` should either:

✅ Run successfully on macOS
**or**
✅ Detect unsupported environments and skip Linux-only steps
**or**
✅ Provide clear documentation stating Linux is required

---

### Suggested Fixes

**Option 1 — OS Detection**
Add a simple OS check in the Makefile:

```bash
uname
```

Then conditionally execute Linux-only commands.

---

**Option 2 — Documentation**
Add a note in the README such as:

> “The Makefile test target currently requires a Linux environment due to networking and GNU utility dependencies.”

---

**Option 3 — Improve Portability (Ideal)**

* Replace `grep -P` with a portable alternative
* Avoid reliance on `docker0`
* Use Docker-native commands where possible

---

### Why This Matters

Improving cross-platform compatibility (or documenting requirements) will:

* Reduce contributor setup time
* Prevent confusion during onboarding
* Improve developer experience

Happy to help test any proposed changes if needed 👍

sydseter:
@prakhar0x01 Thank you for your time. I have assigned you.

prakhar0x01:
@sydseter ,

Closing this, as this is already fixed in the updated code.
Sorry for the inconvenience

Thanks,
@prakhar0x01

PR #2209, opened 2026-02-12
Fix Makefile portability issues on macOS and improve cross-platform contributor experience

prakhar0x01:
Issue Related : #2200

## **Summary**

This PR improves the portability of the Makefile by removing Linux-only assumptions and ensuring that `make test` runs successfully on macOS without failing after the Python test suite completes.

Previously, contributors using macOS encountered errors caused by utilities and networking behavior that exist only on Linux. These changes introduce OS-aware logic and safe fallbacks so the development workflow behaves predictably across platforms.

---

### **Problem**

Running:

```bash
make test
```

on macOS executed all Python tests successfully but failed afterward due to Linux-specific commands.

Errors included:

```
ip: command not found
grep: invalid option -- P
docker0 not found
make: *** docker: No such file or directory
```

This created unnecessary friction for contributors and could negatively impact onboarding.

---

### **Root Cause**

The Makefile relied on behavior specific to Linux environments:

* `ip` command used to fetch the Docker bridge IP
* `docker0` network interface (not exposed on Docker Desktop for macOS)
* `grep -P` which is unsupported by BSD grep
* Hard dependency on a `docker` directory for smoke tests

---

### **Solution**

This PR introduces lightweight OS detection and conditional execution to ensure consistent behavior across platforms.

#### ✔️ OS-Aware HOST_IP Resolution

* Linux continues using the `docker0` interface.
* Non-Linux systems (macOS, Windows) use `host.docker.internal`, which is the Docker-supported hostname.

```make
UNAME_S := $(shell uname -s)

ifeq ($(UNAME_S),Linux)
    HOST_IP ?= $(shell ip addr show dev docker0 | grep -oP "(?<=inet )[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+")
else
    HOST_IP ?= host.docker.internal
    $(info Non-Linux OS detected — using host.docker.internal as HOST_IP)
endif
```

---

#### ✔️ Conditional Docker Smoke Tests

Instead of failing when the `docker` directory is unavailable, the Makefile now safely skips the step:

```make
@if [ -d docker ]; then \
    $(MAKE) -C docker smoke-test; \
else \
    echo "Skipping docker smoke tests (docker directory not found)"; \
fi
```

This preserves existing behavior on Linux while preventing unnecessary failures elsewhere.

---

### **Result**

✅ `make test` now completes successfully on macOS
✅ Python tests and coverage run as expected
✅ Linux workflows remain unchanged
✅ Contributor setup friction is reduced
✅ Developer experience is improved

---

### **Why This Matters**

Improving cross-platform compatibility helps:

* Reduce onboarding time for new contributors
* Prevent confusion caused by environment-specific failures
* Make the project more accessible to the broader developer community
* Align the development workflow with modern multi-OS practices

---

### **Testing**

Verified on:

* **macOS (Apple Silicon)**

  * Python tests: ✅ Passed
  * Coverage threshold: ✅ Passed
  * Makefile execution: ✅ No post-test failure

Linux behavior is preserved because existing commands remain unchanged for Linux environments.

---

### **Backward Compatibility**

No breaking changes.

Linux users will experience identical behavior to before.

---

### **Potential Follow-Ups (Optional Improvements)**

While outside the scope of this PR, future enhancements could include:

* Replacing `grep -P` with a fully portable alternative
* Documenting platform expectations in the README
* Expanding CI to validate macOS workflows

---

### **Conclusion**

This PR focuses on a small but high-impact improvement: making the project easier to build and test across operating systems without altering existing Linux functionality.

Happy to iterate further if maintainers prefer a different approach 🙂


prakhar0x01:
Hi maintainers 👋

It looks like the failing CI check is caused by a vulnerable transitive dependency (`qs`) during the website build.

Since this PR only updates the Makefile for cross-platform compatibility, the failure appears unrelated to the changes introduced here.

Please let me know if you would prefer that I open a separate PR to address the dependency update, or if this is already being tracked elsewhere.

Happy to help!

sydseter:
@prakhar0x01 no that’s fine. There is a patch on the way.

sydseter:
@prakhar0x01 I'll test it out during the weekend when I can run it on my macmini.

PR #2247, opened 2026-02-15
Create webapp-cards-3.0-hi.yaml

SachinAditya:
This PR adds a complete Hindi translation of the WebApp v3.0 cards.

Added:
- source/webapp-cards-3.0-hi.yaml

Details:
- Full translation of all suits, cards, and paragraphs
- Structure, IDs, URLs, and ordering preserved
- English file remains unchanged

This PR only introduces the Hindi translation file and does not modify any existing files.

Please let me know if any formatting or structural adjustments are required.


SachinAditya:
Hi Johan,

The previous PR was closed without merge.

This new PR only adds:
source/webapp-cards-3.0-hi.yaml

No changes were made to the English file.

Please let me know if anything needs adjustment.

Thank you!

sydseter:
@cw-owasp Could you have a look at this?

cw-owasp:
I will take a look

cw-owasp:
Hi @SachinAditya I want to have a look through all of these, but we'll also encourage any other Hindi speaking volunteers to comment too. Hope you don't mind getting more eyes on it!

Also, I had a look at the first card (VE2) and used G.Translate to reverse your text into English. The back translated text is:

_Brian **may** gather information about the underlying configuration, schema, logic, code, software, services, and infrastructure by examining the content of error messages, bad configurations, the presence of default installation files, old, test, backup, or copies of resources, or by exposing source code._

versus the original:

_Brian **can** gather information about the underlying configuration**s**, schema**s**, logic, code, software, services and infrastructure due to the content of error messages, **or** poor configuration, **or** the presence of default installation files **or** old, test, backup or copies of resources, or exposure of source code_

Are the differences which I've highlighted in bold, problems with the back translation, or are the plurals and the words "or" really missing? We use "can" in all the attacks to suggest the ability to do the attack, rather than "may" which almost suggest they are permitted to do so.

cw-owasp:
Can versus May Example 2: G.Translate also suggests VE5 is "may", but it translates VE3 and VE4 as "can" which seem correct. 

Plural example 2: VE5 also seems to have "or the wrong encoding is being used" at the end rather than the plural "or the wrong encoding**s are** being used"

Are these just my G.Translate problems?

SachinAditya:
Thank you for reviewing this carefully.

In Hindi, the phrase “कर सकता है” is used consistently across all attack descriptions to indicate capability (equivalent to “can”), not permission. Google Translate sometimes renders this as “may,” but the intended meaning is “can.”

Regarding plural forms, Hindi does not always explicitly distinguish plural nouns in the same structural way as English. However, the meaning includes plural concepts where the original text uses them.

If needed, I am happy to slightly adjust wording to make the intended “can” meaning clearer in reverse translation.

cw-owasp:
@SachinAditya Thank you very much for explaining. No need to change anything for G.Translate! I will review the rest of the cards now.

One thing that will need changing though is to swap AZ9 and C9, which was _recently changed_ in the EN version:

- AZ9 should be about Michael bypassing access control
- C9 should be about Mike misusing the application

The cards were swapped suits due to feedback during playing the game.

cw-owasp:
I had a look at the rest of the cards, trying to consider G.Translate's failings while doing so. 

In addition to needed to swap AZ9 and C9 (see above), have queries about nine other cards:

- SM7 : I did wonder why both names Graham and Adam are placed adjacent at the start the Hindi. Is that correct?
- CRX : The second part "_or it is not strong enough for the amount of effort the attacker is willing to make_" doesn't seem to be in the Hindi
- C5 : The last part "_, or abuse that trust elsewhere (e.g. in another application)_" doesn't seem to be in the Hindi
- C6, C7, C8, CJ : All of the seem to be missing parts of the EN versions
- CQ : " by the application" is important and seems to be missing from the Hindi
- CK : The denial of service here affects "to some or all of its users" - other users of the same application. Is that explicit in the Hindi?

SachinAditya:
@cw-owasp 
All requested changes have now been applied.
Please let me know if anything further needs adjustment.

cw-owasp:
@SachinAditya Thank you. You have addressed everything I noticed. Maybe we can leave this open a little longer in case anyone else has any feedback.

cw-owasp:
I have another question. Most of the languages use 2-9,10,J,Q,K,A and "Joker" as the Cornucopia cards parameter value displayed on the card faces. But the RU cards use 2-9,10,**В**,**Д**,K,A and "Joker".

So I wondered what the convention is in Hindi for other standard decks of playing cards. Are the EN values normally used on such cards, or is it (like RU) that some of these are a locally-defined character? That could extend to number symbols too.

sydseter:
@cw-owasp good observation. Do you usually use Hindi specific numbers or it is more natural with western numeric symbols?

SachinAditya:
Hi Johan and @sydseter,

In India, standard playing card decks — even in Hindi-speaking regions — normally use Western numeric symbols (2–10) and the English letters J, Q, K, A. It is not common practice to use Devanagari numerals (२, ३, ४…) or Hindi letters for face cards.

So using 2–9, 10, J, Q, K, A and “Joker” is the most natural and culturally accurate convention for Hindi as well.

If you would prefer a localized variant using Devanagari numerals, I’m happy to adjust, but from a real-world usage perspective, the current format matches standard decks used in India.

Best regards,
Sachin

cw-owasp:
Thank you for your guidance. No need to change anything then, we should stick with what's natural: 2–9, 10, J, Q, K, A and “Joker”.

sydseter (approved):

SachinAditya:
Thank you for merging my PR! I really appreciate your time and feedback. It was a great learning experience contributing to this project.

PR #2299, opened 2026-02-19
Add DevOps suit to Companion Edition

mhubl:
Adding the DevOps suit to Companion Edition.

Names from:
 - OWASP SAMM Core Team: https://owaspsamm.org/team/ (Johns deduplicated)
 - Creator of the original SAMM v1.0: https://owaspsamm.org/model/
 - OWASM DSOMM Project Leaders: https://owasp.org/www-project-devsecops-maturity-model/

Note that the DSOMM mappings reference subdimensions, which, as far as I'm aware, are not linkable. Activities, on the other hand, are linkable based on their UUIDs, which is not readable, unless the whole name is also shown. Subdimensions seem like a good fit for a card format due to the small size and relative readability, so that's what I opted for, but the references included on the website will likely have to link to individual activities. Is that okay?

### DIFF
--- a/source/companion-mappings-1.0.yaml
+++ b/source/companion-mappings-1.0.yaml
@@ -384,57 +384,92 @@ suits:
     capec: [ CAPEC-180 ]
     mitre_attack: [ T1190 ]
 -
-  id: SDL
-  name: SSDLC
+  id: DVO
+  name: DevOps
   cards:
   -
-    id: SDL2
+    id: DVO2
     value: 2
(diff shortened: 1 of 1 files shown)


sydseter:
Great! There is a buil error, but it’s not due to the DeOps suite.

cw-owasp:
Really good. I like the orderly A-Z of attacker names too!

cw-owasp (approved):

Issue #2307, opened 2026-02-19
Missing length validation on Player name allows DB error and 500 response


prakhar0x01:
## Description

The `Player` schema currently validates only presence of the `name` field:

**`|> validate_required([:name])`**

However, there is no length validation applied.

The underlying database column is defined as `varchar(255)`.  
When a name longer than 255 characters is submitted:

-   Ecto attempts to insert the record
    
-   PostgreSQL raises `ERROR 22001 (string_data_right_truncation)`
    
-   The LiveView crashes
    
-   The application returns **500 Internal Server Error**
    

This results in improper input validation and unhandled database errors.

## Steps to Reproduce

1.  Navigate to the player creation form.
    
2.  Enter a string longer than 255 characters in the `name` field.
    
3.  Submit the form.
    

## Actual Result

-   Database error: `value too long for type character varying(255)`
    
-   LiveView process crashes
    
-   HTTP 500 response returned
    

## Expected Result

-   Input exceeding allowed length should be rejected at the application layer.
    
-   A validation error message should be shown to the user.

-   No 500 response should be returned.
    

## Impact

-   Application stability issue (user-triggered 500 error)
    
-   Potential minor DoS vector through repeated invalid submissions
    

## Proposed Solution

Add length validation to the `Player` changeset:

**`|> validate_length(:name, min: 1, max: 100)`**

Optionally, enforce a corresponding `maxlength` attribute in the UI for improved UX.

prakhar0x01:
@sydseter ,

Already applied the fix, kindly assign

Thanks,
@prakhar0x01

sydseter:
@prakhar0x01 Hi are you saying there is a bug. I don’t think I remember merging your changes?

prakhar0x01:
Hi @sydseter, 

yes, the bug is still present in the main branch (it causes a 500 error when the name exceeds 255 characters).

<img width="1909" height="691" alt="Image" src="https://github.com/user-attachments/assets/681e6d30-a9e7-4a8b-88a0-daf873ca6da7" />

I have already submitted the fix in PR #2308. Once that is reviewed and merged, it should resolve this issue!

PR #2308, opened 2026-02-19
Fix/player-name-validation : Add length validation to Player name field to prevent DB truncation errors

prakhar0x01:
**Issue Related :** https://github.com/OWASP/cornucopia/issues/2307 

## Summary

This PR adds length validation to the `Player` schema to prevent database-level truncation errors when names exceed the defined column size.

## Changes Made

-   Added `validate_length(:name, min: 1, max: 100)` to `Player.changeset/2`
    
-   Ensures input is validated at the application layer before database insertion
    
-   Prevents `string_data_right_truncation` database error
    
-   Prevents LiveView crash and HTTP 500 response
    

## Before

Submitting a name longer than 255 characters resulted in:

-   PostgreSQL error `22001`
    
-   LiveView crash
    
-   500 response
    

## After

Submitting a name exceeding the allowed limit:

-   Fails validation cleanly
    
-   Returns a changeset error
    
-   Does not trigger a database exception
    
-   No 500 response

### DIFF
--- a/copi.owasp.org/lib/copi/cornucopia/player.ex
+++ b/copi.owasp.org/lib/copi/cornucopia/player.ex
@@ -26,5 +26,6 @@ defmodule Copi.Cornucopia.Player do
     player
     |> cast(attrs, [:name, :game_id])
     |> validate_required([:name])
+    |> validate_length(:name, min: 1, max: 50)
   end
 end


prakhar0x01:
@sydseter ,

Everything is good from my side, 

Please take a look at CI checks.

Thanks,
@prakhar0x01

prakhar0x01:
@sydseter ,

This is done,

Please take a look at : https://github.com/OWASP/cornucopia/issues/2307

Thanks,
@prakhar0x01

cornucopia.owasp.org/package.json
```diff
@@ -18,13 +18,13 @@
 	"devDependencies": {
 		"@sveltejs/adapter-auto": "^7.0.0",
 		"@sveltejs/adapter-cloudflare": "^7.2.6",
-		"@sveltejs/kit": "^2.50.2",
+		"@sveltejs/kit": "^2.52.2",
```
sydseter: Please revert the dependency updates.

sydseter (changes requested):

cornucopia.owasp.org/package.json
prakhar0x01: done

prakhar0x01:
@sydseter , 

I think I need to update the `package.json` , it create unneccessary problems in CI checks.

prakhar0x01:
@sydseter ,

I think we are done, CI checks are quite challenging 😅,

Anyway, all commits are verified, and the changes are applied to the updated code from owasp/master.

Thanks,
@prakhar0x01

cornucopia.owasp.org/src/hooks.server.js
sydseter, line 1: Could this fix go in a separate pull-request?

prakhar0x01: Yeah , It could but then CI checks fails again due to this vulnerability

sydseter (dismissed):

PR #2309, opened 2026-02-19
Ratelimiter

### LINKED ISSUE #1877: Security overhal of Copi
There are various concerns for copi that should get adressed. E.g: making sure that we maintain availability under an attack. I know a couple of patterns that would work like [(CAPEC 212, functionality missues)](https://capec.mitre.org/data/definitions/212.html). This, we should take care of. We won’t be able to remove the threat completely, but we should minimize it. Knowledge og Elixir requiered. The person that fixes this deserves to be mentioned in the OWASP Cornucopia Hall of Fame for sure.

How? We need to put a limit on the numer of users.

### LINKED ISSUE #2307: Missing length validation on Player name allows DB error and 500 response

sydseter:
In this pull-request:

- This will add rate limiting for creating games and players when the request comes from the same ip.
- Fixes #1877 closes #2307


sydseter:
npm audit is down which is why the website build fails. The website isn’t touched here.

PR #2319, opened 2026-02-20
Privacy

sydseter:
Adding privacy notice to Copi


sydseter:
The build fails because black complains about a formating style issue in one of the converter scripts. it’s a pain to fix given that it can only be done on a linux machine where I can run with elevated privileges in order to execute python.

sydseter:
There are some vulnerabilities on the website as well that cause the build to fail. Those, as well, have to be looked at separately in case something breaks.

sydseter:
The coverage on Copi has gone a bit down due to the work on the rate limiter and this privacy notice. There should be a separate task to get it above 80 again. I have prioritized getting the privacy and security part done. It’s has decreased the test coverage 5%, but test coverage is still up from 45% since January.

PR #2326, opened 2026-02-20
Ensure the route to api goes to api/docs

sydseter:


prakhar0x01:
@sydseter ,

Also add  this

https://cornucopia.owasp.org/api/cre/webapp
https://cornucopia.owasp.org/api/cre/mobileapp

Thanks,
@prakhar0x01

sydseter:
all good.

PR #2339, opened 2026-02-21
fix: use gh release upload for ZAP report upload to pre-release

### LINKED ISSUE #2333: Zap proxy is not uploading the reports to the  pre-release
We need to make sure we upload the zap report to the pre-release during the nighttime. Using the [softprops github action](https://github.com/OWASP/cornucopia/blob/52de40c4785f8c41281b0f40cf9981c1194dd2f8/.github/workflows/zap-nightly-scan.yml#L142) is not a good idea because it won’t upload files to an existing release.
The right way is to do what we do for generating the sbom: 
https://github.com/OWASP/cornucopia/pull/2331

Mysterio-17:
### Description
The softprops/action-gh-release action cannot upload files to an already existing release, so the nightly ZAP DAST reports were silently failing to attach to the pre-release. This replaces it with gh release upload --clobber, which properly uploads files to existing releases.

Closes #2333 

### Changes

- Replaced softprops/action-gh-release with gh release upload "pre-release" --clobber
- Each report file is checked for existence before upload (since ZAP scan uses || true)
- Uses GITHUB_TOKEN for authentication (already available via permissions: contents: write)

### DIFF
--- a/.github/workflows/zap-nightly-scan.yml
+++ b/.github/workflows/zap-nightly-scan.yml
@@ -141,13 +141,15 @@ jobs:
 
       - name: Upload reports to pre-release
         if: always()
-        uses: softprops/action-gh-release@5be0e66d93ac7ed76da52eca8bb058f665c3a5fe # v2.4.2
-        with:
-          tag_name: pre-release
-          prerelease: true
-          name: Latest pre-release
-          files: |
-            zap-reports/copi_dast_report.html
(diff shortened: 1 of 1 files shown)


Mysterio-17:
Hello @sydseter , I've replaced `softprops/action-gh-release` with `gh release upload --clobber` so ZAP reports are properly uploaded to the existing pre-release.
Kindly have a look at it.

PR #2426, opened 2026-02-27
pt_br translation of 43.0 cards

izar:
Hi there, 

1. used a LLM to do a first pass translation
2. compared translation and EN using cosine distance - anything lower than a 0.7 got translated by hand
3. did a reverse translation and comparison with the original for further verification
4. verified that the YAML structure remained identical to the EN version
5. verified the whole file by hand for consistency and tone


sydseter:
@izar Brilliant! Thanks so much for the help!

PR #2428, opened 2026-02-27
Adding Dutch translater to the list

sydseter:


sydseter:
Also adding Izar Tarandach for the Brazillian.

PR #2440, opened 2026-02-28
Add link to privacy notice

sydseter:
In this pull-request:

Adding link to privacy notice when adding  player names


sydseter:
the build fails because the test coverage on copi is lower then 80 %. I am working on increasing it.

Issue #2447, opened 2026-02-28
Optimize CI performance: Add pnpm store caching to website build & deploy workflows for faster CI runs


khushal-winner:
**Is your feature request related to a problem? Please describe.**  
The current GitHub Actions workflows for the website (`cornucopia.owasp.org`) — including `build-website.yml`, `build-website-staging.yml`, `deploy-website-production.yml`, and `deploy-website-staging.yml` — run `pnpm install` from scratch on every execution. This reinstalls thousands of dependencies (pnpm-lock.yaml has thousands of lines/entries) repeatedly, leading to unnecessarily slow build times, higher network usage, increased CI minutes consumption, and slower feedback loops for contributors working on the SvelteKit frontend, scripts, or YAML content. This is especially noticeable during frequent PR checks, staging deploys, and release workflows.

**Describe the solution you'd like**  
Implement standard GitHub Actions caching for the pnpm store directory in all relevant website-related workflows. This uses `actions/cache@v4` with a cache key based on the hash of `**/pnpm-lock.yaml` (and optionally the OS runner), which is the recommended pattern for pnpm projects.

Proposed implementation (to be added after the `Setup pnpm` / `pnpm setup` step and before `pnpm install` in each affected workflow):

```yaml
- name: Get pnpm store directory
  id: pnpm-cache
  shell: bash
  run: |
    echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_OUTPUT

- name: Setup pnpm cache
  uses: actions/cache@v4
  with:
    path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
    key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
    restore-keys: |
      ${{ runner.os }}-pnpm-store-
```

This caches the global pnpm store (~/.pnpm-store or equivalent), so subsequent runs only download new/changed packages. After the cache is restored, `pnpm install` becomes near-instant for unchanged lockfiles. No changes are needed to package.json, pnpm-lock.yaml, or any application code — it's purely a CI optimization.

This will be applied to:
- `.github/workflows/build-website.yml`
- `.github/workflows/build-website-staging.yml`
- `.github/workflows/deploy-website-production.yml`
- `.github/workflows/deploy-website-staging.yml`

(And any other workflow that runs `pnpm install` in the `cornucopia.owasp.org/` context.)

**Describe alternatives you've considered**  
[...]

PR #2448, opened 2026-02-28
feat: implement pnpm caching in CI workflows

khushal-winner:
**Title:** Optimize CI performance: Add pnpm store caching to all website build & deploy workflows

**Closes:** #2447

**Description:**

This PR implements pnpm store caching in the four website-related GitHub Actions workflows to significantly speed up `pnpm install` steps, reduce network usage, lower CI minutes consumption, and provide faster feedback for contributors (especially useful during active development, PR checks, staging deploys, and GSoC contributions to `cornucopia.owasp.org`).

**Changes:**

Added the following steps **before** each `pnpm install` in all affected workflows:

```yaml
- name: Get pnpm store directory
  id: pnpm-cache
  shell: bash
  run: |
    echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_OUTPUT

- name: Setup pnpm cache
  uses: actions/cache@v4
  with:
    path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
    key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
    restore-keys: |
      ${{ runner.os }}-pnpm-store-
```

Updated workflows:
- `.github/workflows/build-website.yml`
- `.github/workflows/build-website-staging.yml`
- `.github/workflows/deploy-website-production.yml`
- `.github/workflows/deploy-website-staging.yml`

**Benefits:**
- **Faster CI runs** — `pnpm install` becomes near-instant on cache hit (after initial build)
- **Cache invalidation** handled correctly via `pnpm-lock.yaml` hash
- **Zero risk** — cache miss falls back to normal install behavior
- **Follows official best practices** — based on [pnpm CI guide for GitHub Actions](https://pnpm.io/continuous-integration#github-actions) and avoids less reliable `node_modules` caching

**Validation:**
- YAML syntax verified in all files
- No changes to application code, `package.json`, or `pnpm-lock.yaml`
- Backward compatible and low-risk

This should improve developer experience and CI efficiency for the SvelteKit-based website.

### DIFF
--- a/.github/workflows/build-website-staging.yml
+++ b/.github/workflows/build-website-staging.yml
@@ -36,6 +36,21 @@
             uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
             with:
               node-version: 20.18.2
+          
+          - name: Get pnpm store directory
+            shell: bash
+            run: |
+              echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_OUTPUT
+            id: pnpm-cache
+          
+          - uses: actions/cache@v4
+            name: Setup pnpm cache
+            with:
+              path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
+              key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+              restore-keys: |
+                ${{ runner.os }}-pnpm-store-
+          
           - name: Build
             working-directory: cornucopia.owasp.org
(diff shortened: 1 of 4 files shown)


sydseter:
I checked whether this makes the build faster, it doesn’t. The build took around 1.26 min before, now it takes 1.32 

It’s 5 seconds slower. The cache step takes time so we end up not saving any. The build is fast enough I think. Let’s close this.

PR #2407, opened 2026-02-26
fix: add dynamic route for /api/cre/webapp and /api/cre/mobileapp

Mysterio-17:
## Summary
Fixes : #2380 

Creates a dynamic `[edition]/+server.ts` route handler that serves edition metadata (supported languages, version, edition name) for any valid edition, replacing the need for separate hardcoded route files per edition.

A post-build script is also added to generate the static edition metadata JSON files during the build, since SvelteKit's `adapter-static` cannot prerender both `/api/cre/webapp` (file) and `/api/cre/webapp/en` (directory) at the same path without a filesystem conflict.

## Changes

- Added `src/routes/api/cre/[edition]/+server.ts` - Dynamic route handler serving edition metadata using DeckService.
- Added `script/generate-edition-meta.js` - Post-build script generating static JSON for production builds.
- Modified `package.json` - Added `generate-edition-meta.js` to build, build-stage, and productionbuild scripts.
- Deleted `src/routes/api/cre/webapp/+server.js` - Redundant, replaced by the dynamic [edition] route. Can be reverted if preferred.
- Deleted `src/routes/api/cre/mobileapp/+server.js` - Redundant, replaced by the dynamic [edition] route. Can be reverted if preferred.

## Testing

- All existing tests pass (npx vitest run)
- npm run build succeeds

PR #2450, opened 2026-02-28
fix: add Content-Type header to CRE API response

Mysterio-17:
## Problem
After #2407 was merged, the `/api/cre/webapp` and `/api/cre/mobileapp` routes were working but missing the explicit `Content-Type: application/json` header, which could cause issues for API consumers.

## Solution
Added explicit `Content-Type: application/json` header to the JSON response in the `[edition]/+server.ts` route handler, following the same pattern used in the working `[edition]/[lang]/+server.ts` route.

## Changes
- Modified `cornucopia.owasp.org/src/routes/api/cre/[edition]/+server.ts`
- Added headers object to the `json()` response with explicit `Content-Type: application/json`

## Related
Follow-up to #2407


Mysterio-17:
Hello @sydseter , I've fixed the follow-up issue by adding the explicit Content-Type: application/json header and updated the PR description as suggested by Copilot to clarify this is a follow-up to #2407, not the original 404 fix.

Issue #2488, opened 2026-03-03
Bug: Unhandled KeyError on missing 'Categories' key in capec_map_enricher.py


immortal71:
## Describe the bug

In `scripts/capec_map_enricher.py`, the `extract_capec_names()` function correctly guards against missing top-level keys (`Attack_Pattern_Catalog`, `Attack_Patterns`, `Attack_Pattern`) but then directly accesses `catalog["Categories"]["Category"]` without any guard check. If the CAPEC JSON data has no `"Categories"` key, the script crashes immediately with an unhandled `KeyError`.

## Affected file

`scripts/capec_map_enricher.py` — `extract_capec_names()` function

## Code snippet (problematic section)

```python
# After all the earlier guards for Attack_Pattern...
# Lines ~62-67: NO guard for "Categories" key
categories = catalog["Categories"]["Category"]  # KeyError if "Categories" is absent
for category in categories:
    if "_ID" in category and "_Name" in category:
        capec_id = int(category["_ID"])
        capec_name = category["_Name"]
        capec_names[capec_id] = capec_name
```

## Expected behavior

The function should check for the existence of `"Categories"` and `"Category"` before accessing them, consistent with the defensive guards already applied to `Attack_Pattern_Catalog`, `Attack_Patterns`, and `Attack_Pattern` earlier in the same function. A warning should be logged if absent rather than raising an unhandled exception.

## Proposed fix

```python
if "Categories" not in catalog:
    logging.warning("No 'Categories' key found in catalog")
else:
    categories_section = catalog["Categories"]
    if "Category" not in categories_section:
        logging.warning("No 'Category' key found in categories section")
    elif not isinstance(categories_section["Category"], list):
        logging.warning("'Category' is not a list")
    else:
        for category in categories_section["Category"]:
            if "_ID" in category and "_Name" in category:
                capec_id = int(category["_ID"])
                capec_name = category["_Name"]
                capec_names[capec_id] = capec_name
```

## Steps to reproduce

1. Provide a `3000.json` where `Attack_Pattern_Catalog` has no `"Categories"` key
2. Run: `python scripts/capec_map_enricher.py`
3. Observe unhandled `KeyError: 'Categories'`

## Additional context

This is inconsistent with the existing defensive coding style used for all preceding key accesses in the same function. The fix should follow the same pattern already established.

immortal71:
@sydseter Is this good to work on, I have complete previous pr which I was working on !!, is this good to go ?

PR #2506, opened 2026-03-04
fix: guard against missing Categories key in extract_capec_names

immortal71:
**Issue:** #2488

## Problem
extract_capec_names() in scripts/capec_map_enricher.py accessed catalog['Categories']['Category'] directly without any guard, causing an unhandled KeyError crash if the CAPEC JSON has no Categories section.

## Fix
Added defensive checks matching the existing guard style used for Attack_Patterns and Attack_Pattern earlier in the same function. If Categories or Category is absent/malformed, a warning is logged and the block is skipped — the function continues and returns whatever attack pattern names were already extracted.


immortal71:
@sydseter  is this good to go ? can u review this pr !!

scripts/capec_map_enricher.py
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 # This script takes JSON from
+# Minor non-functional comment to create a small, safe diff for PR creation (Issue #2487)
```
sydseter: Please remove this comment.

scripts/capec_map_enricher.py
```diff
@@ -1,5 +1,7 @@
...
+# No behavioral changes - purely a comment to allow PR creation when branch has no other diffs.
```
sydseter: And remove this.

copi.owasp.org/coveralls.json
sydseter, line 1: revert the changes in this file.

sydseter:
@immortal71 There are some conflicts and I have notes some things in the code that should be removed.

immortal71:
@sydseter  okay !!

immortal71:
@sydseter  is this good to now ???

Issue #2502, opened 2026-03-04
🚨 Race condition in game with actual video


khushal-winner:
- **Describe the bug**  
  A race condition occurs when multiple users (or the same user with rapid clicks) trigger the "Start Game" action at nearly the same time. Because card dealing and the `started_at` timestamp update are not wrapped in a transaction — and there is no unique constraint on `DealtCard` — both requests pass the `game.started_at == nil` check and then both execute the full card-dealing loop. This results in every card from the deck being inserted twice.  
  
  **Observed behavior**  
- Normal (single click): each player receives exactly 26 unique cards (correct for a 3-player game)
- Race condition triggered (two near-simultaneous "Start Game" clicks): each player receives 52 cards (the entire deck is dealt twice)

**Before**

https://github.com/user-attachments/assets/5f3977f9-92fd-47d9-9d57-685b8ede64d0

**After** (triggered the start game same time with mobile and laptop , mobile is not shown in the video because this is a screen recording)

https://github.com/user-attachments/assets/e84302a9-07bb-42e0-9654-ba7d9b644735

  
  **Expected behavior**  
  Clicking "Start Game" should deal each card exactly once, mark the game as started (`started_at` set), and prevent any concurrent dealing process from duplicating cards.  
  
  **Steps to reproduce**  
	1. Create/join a game with >2 players
	2. Ensure game is ready to start (both players present, deck loaded)
	3. Have two browser tabs (or two users) click "Start Game" within ~100–300 ms of each other
	4. Check the dealt cards — each player will have 52 cards instead of 26  
  
**Desktop environment**  
OS: Windows  
Browser: Chrome  
Version: 130+ (also reproducible in Edge/Firefox)  
  
**Additional context**

- Root cause: `Repo.insert!` calls in `show.ex` (lines ~77–84) run outside a transaction
- No unique index exists on `dealt_cards (card_id)` or `(card_id, player_id)`
- The `started_at` nil check is not atomic with the inserts
- Fix suggestion: wrap both the card inserts and `update_game` call in `Repo.transaction/1` (as previously shown)

khushal-winner:
@sydseter , found a very interesting bug, would love to work on this ASAP :))

sydseter:
@khushal-winner great, I have assigned you.

khushal-winner:
@sydseter , seperately created tests and fix, to keep the changes islolated and for better maintainability for me as well as the maintainers

sydseter:
[...]

PR #2509, opened 2026-03-04
Fix race condition in game start with atomic transactions

khushal-winner:
Closes - #2502 

- Implement atomic database transactions with row-level locking
- Fix stale data usage by preloading fresh data within transaction
- Add comprehensive error handling for race conditions
- Create data cleanup migration to remove existing duplicates
- Add unique constraints to prevent future race conditions
- Enhance broadcast handling to occur after transaction completion

Resolves duplicate card dealing bug when multiple users click 'Start Game' simultaneously.

- before (52 cards to each with race condition persists)

https://github.com/user-attachments/assets/d0e59898-6c96-40f8-a1dd-0d21f654fb82

- after (26 cards to each with race condition fixed)

https://github.com/user-attachments/assets/7ac75ab4-dba0-4ab2-86d5-7f9604200f94


copi.owasp.org/priv/repo/migrations/20250304181000_cleanup_duplicate_dealt_cards.exs
sydseter, line 1: This is unnecceary. No cleanup is needed. It carries with it risks.

sydseter:
I am a bit sceptical to this. I did some tests and the game is behaving strange when I use your transaction implementation. I am getting errors through the API. It's may not be an issue, but these errors weren't there before. I would like to avoid getting new issues.

The race condition is not really an issue since there won't be more then one person starting the game. So I think I want to close this issue.
Skjermbilde 2026-03-18 001647

Issue #2507, opened 2026-03-04
Standardize Language Codes in check_translations.py


abhijit9040:
Problem
Language codes in translation filenames use underscores (e.g., no_nb, pt_br), but the lang_names mapping in scripts/check_translations.py uses hyphens (no-nb, pt-br).

Because of this mismatch, the translation report shows raw codes like ## no_nb instead of readable names like ## Norwegian.

abhijit9040:
Hi @sydseter , please have a review ,can you please assign this issue ?

PR #2524, opened 2026-03-05
fix: standardize language codes in check_translations.py 

abhijit9040:
**Description**

This PR updates the language code mapping in scripts/check_translations.py to match the naming convention used in translation YAML files.

**Problem**

Translation files use underscores for regional codes (e.g., no_nb, pt_br), while the script expected hyphenated codes (no-nb, pt-br). This caused the generated translation_check_report.md to display raw language codes instead of readable names. Hindi (hi) was also missing from the mapping.

Changes

- Updated language mappings to use underscore format (no_nb, pt_br, pt_pt)
- Added "hi": "Hindi" to support the Hindi translation

Result

The translation report now shows correct language names (e.g., Norwegian, Portuguese (Brazil), Hindi) instead of raw codes.

reference issue :- #2507

### DIFF
--- a/scripts/check_translations.py
+++ b/scripts/check_translations.py
@@ -159,6 +159,7 @@ class TranslationChecker:
         lang_names = {
             "es": "Spanish",
             "fr": "French",
+            "hi": "Hindi",
             "hu": "Hungarian",
             "it": "Italian",
             "nl": "Dutch",


abhijit9040:
Hi @sydseter  @rewtd , let me know if any further changes are needed.

sydseter:
Part of this is already taken care of. You can add the Hindi translation if you want.

abhijit9040:
Hi @sydseter, thanks for the feedback!
I checked the upstream master branch, and the lang_names dictionary in scripts/check_translations.py still lists the hyphenated codes ("no-nb", "pt-br", "pt-pt"). Could you clarify which part has already been addressed so I can scope my PR correctly?
For Hindi, I’ve added "hi": "Hindi" as suggested. Please let me know if you’d prefer this PR to focus only on that addition, and I’ll adjust accordingly.

sydseter (approved):

Issue #2539, opened 2026-03-06
Scores missing from player pages after game ends


khushal-winner:
**Describe the bug**  
when game ends Scores are only visible on the game page (`/games/:id`), not on individual player pages (`/games/:id/players/:player_id`). Players cannot see scores (their own or others') from their personal view.

**Expected behavior**  
Scores (individual + total, possibly leaderboard/ranking) should be shown on player-specific pages so each player can view results without leaving their view.

**Video Demo**

https://github.com/user-attachments/assets/82b34c10-8088-420b-aa92-ecbbcf7b56dd

https://github.com/user-attachments/assets/7f45901c-2066-463c-92b6-8a1310104417

**Desktop**  
- OS: any
- Browser: any
- Version: any

**Additional context**  
The game ends with `finished_at` set and scores calculated (card majority + highest card bonus), but they're hidden from player routes. This limits visibility during/after play — players must know/use the top-level game URL to see results, which feels inconsistent for a multi-player experience. Adding score display (at least own score) to player LiveView would improve usability.

khushal-winner:
@rewtd , take a look at this

rewtd:
Perhaps players should simply be routed to the game page at the end?

khushal-winner:
> Perhaps players should simply be routed to the game page at the end?

@rewtd , if we redirect to the game page when game ends but when the player tries to access his own player page again (`https://copi.owasp.org/games/:game_id/players/:player_id`) it should also shows the score report.

we can do one of these

1. add score table on player page too & not redirect anything
2. always redirects the player page to the game page once the game ends making the player page unaccessible anymore

rewtd:
As the player page no longer has any value once the game as ended I think we go with option 2.

khushal-winner:
> As the player page no longer has any value once the game as ended I think we go with option 2.

@rewtd , sure assign me i'll fix it

sydseter:
I think we go with the approach where the user have a link they can click that takes them to the game's overview since that has already been implemented.

sydseter:
Issue #2539 does not need further handling under the agreed approach.

Current behavior in copi.owasp.org/lib/copi_web/live/player_live/show.html.heex:

• Finished games show a “Go to game summary” link.

• The game summary page displays the scores.

• The finished player view no longer exposes gameplay controls.

PR #2544, opened 2026-03-06
feat: redirect players to game page when game ends

khushal-winner:
Closes - #2539 

- Add game_finished?/1 function to Game module
- Implement game-end redirect in PlayerLive.Show
- Redirect from /games/:game_id/players/:player_id to /games/:game_id
- Handle both initial access and real-time game end events
- Use push_navigate for seamless LiveView redirects

https://github.com/user-attachments/assets/0fcdff17-3c26-4fab-a742-d84acce9a2e7

Fixes issue where players could access finished game pages instead of being redirected to game summary.

### DIFF
--- a/copi.owasp.org/lib/copi/cornucopia/game.ex
+++ b/copi.owasp.org/lib/copi/cornucopia/game.ex
@@ -62,4 +62,8 @@ defmodule Copi.Cornucopia.Game do
 
     Enum.count(players_still_to_play) > 0
   end
+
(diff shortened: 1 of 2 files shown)


sydseter:
I’ll need to test this out, but it looks good.

sydseter:
Actually, the point is that you should have a link published at the end that you can click to go to the overview, not an automatic redirect like here. I believe the link should be working correctly now.

PR #2551, opened 2026-03-07
fix: use exact equality for ASVS shortcode matching in convert_capec.py

### LINKED ISSUE #2547: Bug: `shortcode in subitem[\"Shortcode\"]` substring match in convert_capec.py generates wrong ASVS links
## Describe the bug

In `scripts/convert_capec.py`, the `createlink()` function uses Python's `in` operator (substring containment) instead of `==` (equality) when matching ASVS shortcodes. This silently returns a link pointing to the **wrong ASVS requirement** whenever a shorter shortcode is a substring of a longer one.

## Affected file

`scripts/convert_capec.py` — `createlink()` function (line 127)

## Code snippet (problematic section)

```python
def createlink(data: dict[str, Any], shortcode: str, asvs_version: str) -> str:
    for i in data["Requirements"]:
        name = str(i["Ordinal"]).rjust(2, "0") + "-" + ...
        for item in i["Items"]:
            itemname = ...
            for subitem in item["Items"]:
                if shortcode in subitem["Shortcode"]:   # ← BUG: substring check, not equality
                    return f"[{shortcode}](/taxonomy/asvs-{asvs_version}/{name}/{itemname}#{subitem['Shortcode']})"
    return shortcode if shortcode else ""
```

## Why it's a bug

Python's `in` operator checks **substring containment**, not equality:

```python
"V1.1.1" in "V1.1.15"   # True  — WRONG match
"V1.1.1" in "V1.1.1"    # True  — correct match
```

Because `createlink` returns on the **first** match, if the ASVS data contains `V1.1.15` before `V1.1.1` during iteration, then looking up shortcode `"V1.1.1"` returns a link anchored to `V1.1.15` instead — silently, with no error or warning.

ASVS 5.0 has many multi-digit requirement ordinals (e.g. `V2.1.10`, `V2.1.11`, `V6.2.10`) where this prefix-collision is highly likely.

## Expected behavior

The comparison should use equality, so only the exact matching requirement produces a link:

```python
if shortcode == subitem["Shortcode"]:
```

## Steps to reproduce

1. Run `python scripts/convert_capec.py`
2. Inspect generated `index.md` files under `cornucopia.owasp.org/data/taxonomy/en/capec-3.9/`
3. Find a CAPEC entry whose ASVS mapping includes a shortcode like `V2.1.1`
4. Observe that the generated link may anchor to `V2.1.10` or `V2.1.11` instead of `V2.1.1`

## Proposed fix

```python
# Before (line 127):
if shortcode in subitem["Shortcode"]:

# After:
if shortcode == subitem["Shortcode"]:
```

## Additional context

- This bug produces incorrect ASVS documentation links with no runtime error, making it hard to detect without manually auditing generated pages.
- The bug is consistent: it always returns the first requirement whose code contains the target as a substring, which may or may not be the correct one depending on data ordering.
- The `createlink_list()` function that calls `createlink()` is used to generate Related ASVS links in CAPEC taxonomy pages on cornucopia.owasp.org.

- [X] Yes
- [ ] No

immortal71:
Fixes #2547.

## Problem

`createlink()` in `scripts/convert_capec.py` used Python's `in` operator to compare ASVS shortcodes:

```python
if shortcode in subitem["Shortcode"]:
```

`in` checks substring containment, not equality. So `"V1.1.1" in "V1.1.10"` is `True`. When ASVS data contains `V1.1.10` before `V1.1.1` in iteration order, looking up `V1.1.1` returns a link anchored to `V1.1.10` — silently wrong output with no error.

## Fix

Change to exact equality:

```python
if shortcode == subitem["Shortcode"]:
```

## Tests added

- `test_createlink_no_prefix_collision` — V1.1.1 must not match V1.1.10/V1.1.11 when they appear first
- `test_createlink_exact_match_multi_digit` — V1.1.10 must match exactly, not V1.1.1
- `test_createlink_empty_shortcode` — empty shortcode returns empty string

All 53 tests in `convert_capec_utest.py` pass.

### DIFF
--- a/copi.owasp.org/test/copi/cornucopia_logic_test.exs
+++ b/copi.owasp.org/test/copi/cornucopia_logic_test.exs
@@ -96,7 +96,7 @@ defmodule Copi.CornucopiaLogicTest do
     create_card("Wild Card", "1")
     create_card("Hearts", "5")
     create_card("WILD CARD", "2")
-    
+
     suits = Cornucopia.get_suits_from_selected_deck("webapp")
     
     refute "Wild Card" in suits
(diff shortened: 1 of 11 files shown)


immortal71:
@sydseter  is this good to go ??? is commentpr issues solved >>

Suresh-Krishna-P:
Hello @immortal71, The issue was that using Python's 'in' operator for shortcode matching caused prefix collisions - for example, looking up V1.1.1 would incorrectly match V1.1.10 if it appeared first in the iteration order. The fix is simple but effective: changing to exact equality comparison with ==. The addition of comprehensive regression tests demonstrates good defensive programming practices and ensures this type of bug won't reoccur. The PR is clean, targeted, and addresses the specific issue without unnecessary changes. Keep Going.

immortal71:
@Suresh-Krishna-P  hey not to sound rude but i am curious r u maintainer of memberof the org ,if is that so then Can you review the pr I think its ready for the review

Suresh-Krishna-P:
No, I'm not a maintainer.

PR #2552, opened 2026-03-07
fix: pin outdated GitHub Actions versions in CI workflows

Adarshkumar0509:
Fixes #2542

Changes:
- Updated smoke-tests.yaml: actions/checkout 
  v6.0.1 → v6.0.2
- Updated sbom-generate.yml: actions/checkout 
  @v4 → v6.0.2 (pinned)
- Updated sbom-generate.yml: actions/setup-node 
  @v4 → v6.2.0 (pinned)

Note: run-tests-generate-output.yaml was already 
updated to v6.0.2 in master.

PR #2576, opened 2026-03-09
fix: handle nil player in play_card API to prevent 500 crash

### LINKED ISSUE #2562: Bug: API play_card crashes with 500 error when player_id doesn't belong to game_id
**Describe the bug**
In api_controller.ex, when a valid game_id is provided but the player_id doesn't belong to that game, Enum.find returns nil for player. The code then immediately tries to access player.dealt_cards on this nil value before the if player && dealt_card safety check, causing a 500 Internal Server Error crash.

**To Reproduce**
1. Get a valid game_id from any active game
2. Use a player_id that does not belong to that game
3. Send a PUT request to /api/games/{game_id}/players/{invalid_player_id}/card with a dealt_card_id
4. Server crashes with 500 Internal Server Error
5. (curl -X PUT https://copi.owasp.org/api/games/VALID_GAME_ID/players/INVALID_PLAYER_ID/card \
  -H "Content-Type: application/json" \
  -d '{"dealt_card_id": "1"}')

**Expected behavior**
The API should return a 404 Not Found response when the player_id does not belong to the game, instead of crashing.

**Additional context:**
Vulnerable code in code in lib/copi_web/controllers/api_controller.ex:
<pre>
elixirplayer = Enum.find(game.players, fn player -> player.id == player_id end)
dealt_card = Enum.find(player.dealt_cards, ...) # crashes if player is nil
</pre>
**Fix:**
<pre>
elixirif player do
  dealt_card = Enum.find(player.dealt_cards, ...)
else
  conn |> put_status(:not_found) |> json(%{"error" => "Player not found in game"})
end
</pre>

- [x] Yes
- [ ] No

Adarshkumar0509:
Fixes #2562

## Problem
When a valid game_id is provided but the player_id doesn't belong to that game, 
Enum.find returns nil for player. The code then immediately tries to access 
player.dealt_cards on this nil value, causing a 500 Internal Server Error crash.

## Fix
Added a nil check for player before accessing player.dealt_cards. If player is 
not found in the game, the API now returns a proper 404 response instead of crashing.

## Test
Added a test case to api_controller_test.exs that verifies the API returns 404 
when player_id doesn't belong to the game.

### DIFF
--- a/copi.owasp.org/lib/copi_web/controllers/api_controller.ex
+++ b/copi.owasp.org/lib/copi_web/controllers/api_controller.ex
@@ -1,47 +1,43 @@
 defmodule CopiWeb.ApiController do
   use CopiWeb, :controller
   alias Copi.Cornucopia.Game
-
-
   def play_card(conn, %{"game_id" => game_id, "player_id" => player_id, "dealt_card_id" => dealt_card_id}) do
     with {:ok, game} <- Game.find(game_id) do
-
       player = Enum.find(game.players, fn player -> player.id == player_id end)
-      dealt_card = Enum.find(player.dealt_cards, fn dealt_card -> Integer.to_string(dealt_card.id) == dealt_card_id end)
-
-      if player && dealt_card do
-        current_round = game.rounds_played + 1
-
-        cond do
-          dealt_card.played_in_round ->
-            conn |> put_status(:not_acceptable) |> json(%{"error" => "Card already played"})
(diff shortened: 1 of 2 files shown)


Adarshkumar0509:
@sydseter @rewtd this PR fixes #2562 the API was crashing with a 500 error when player_id doesn't belong to the game. I've also added a test case to cover this scenario. The failing check is the pre-existing coverage issue (80.8% vs 85% threshold) unrelated to these changes, as acknowledged in PR #2552."

Issue #2233, opened 2026-02-14
Add version drop down for the individual card pages


sydseter:
Similarly as we do today with the language drop down menu on e.g: https://cornucopia.owasp.org/edition/webapp/VE2/2.2/en

The drop down menu should be visible on the following page controllers (for all possible card selection options) :

https://cornucopia.owasp.org/edition/webapp/VE2/2.2/en
https://cornucopia.owasp.org/edition/webapp/VE2/2.2
https://cornucopia.owasp.org/edition/webapp/VE2

https://cornucopia.owasp.org/edition/mobileapp/PC2/1.1/en
https://cornucopia.owasp.org/edition/mobileapp/PC2/1.1
https://cornucopia.owasp.org/edition/mobileapp/PC2

https://cornucopia.owasp.org/cards/VE2
https://cornucopia.owasp.org/cards/PC2

We need a way to show other versions of the card. By default, for e.g: the Website App edition should display version: v2.2 and language: English, but with the possibility to switch to other versions like v3.0. If displaying the spanish 2.2 version and switching to the 3.0 version, if 3.0 doesn’t have a Spanish translation, show the English version as the default.

Display the language and version drop down meny side-by-side, but order then underneath each other on mobile, like in the image below.

![Image](https://github.com/user-attachments/assets/f7d11885-584c-41b0-807d-f7bc356731c8)

prakhar0x01:
@sydseter ,

I want to work on it.

### I’ve reviewed the requirements and here’s how I propose we can implement it:

##### Language & Version Dropdowns

- Extend the existing language dropdown component to include a version selector side-by-side on page.
- Populate versions dynamically from available editions (v2.2, v3.0, etc.).

##### Fallback Logic

- When switching versions, check if the selected language exists.
- If not, fallback to English by default.

##### Page Coverage

- Ensure the dropdowns appear on all relevant page controllers:

`/edition/webapp/{card}/{version}/{lang}`

`/edition/webapp/{card}/{version}`

`/edition/webapp/{card}`

`/cards/{card}`

##### Navigation / UX

- Switching language or version should redirect users to the correct page while maintaining card context.
- Maintain consistent styling with the current dropdown.

I can create a PR implementing this logic with responsive styling, dynamic version/language handling, and graceful fallback.

Would you like me to go ahead with this approach?

sydseter:
@prakhar0x01 I have assigned you. Thank you for your time.

aadilrizwan:
[...]

PR #2588, opened 2026-03-10
Fix: version dropdown loading on card pages

prakhar0x01:
# Issue Resolved : https://github.com/OWASP/cornucopia/issues/2233

## Summary

This PR fixes an issue where the **Version dropdown on individual card pages was not populated correctly**.

Previously, the server-side loader attempted to retrieve available versions without passing the required `edition` argument to `DeckService.getVersions()`. Because the method requires an edition parameter, this resulted in the version list being undefined, causing the **Version selector to render empty in the UI**.

This change ensures that the correct edition is used when retrieving the list of available versions.

----------

## Root Cause

`DeckService.getVersions()` is defined as:

public static getVersions(edition: string): string[]

However, the loader previously called it without providing the required edition value, resulting in no versions being returned.

----------

## Fix

The loader now retrieves the edition from the resolved card object and passes it to `DeckService.getVersions()`.

### Before

const  versions  =  DeckService.getVersions();

### After

```js
const  edition  =  card.edition;  
const  versions  =  DeckService.getVersions(edition);
```

----------

## Impact

This change restores the expected functionality of the **Version selector on card pages**, ensuring that users can properly view and navigate between available deck versions for the selected edition.

Affected UI component:

`/cards/[card]`

Specifically:

Home > cards > VE2  
Version: [dropdown]

The dropdown now correctly displays versions such as:

2.2  
3.0

for the `webapp` edition.

----------

## Testing

Tested locally by:

1.  Running the development server
    
2.  Navigating to several card pages such as:
    

/cards/VE2  
/cards/DVE9  
/cards/DVEX

3.  Confirming that the **Version dropdown is now populated correctly**.


prakhar0x01:
@sydseter ,

Sorry for the delay, Please take a look.

Thanks,
@prakhar0x01

Issue #2587, opened 2026-03-10
Archive Directory Inclusion in YAML File Collection


khushal-winner:
**Description:**  
The `get_files_from_of_type()` function in `scripts/convert.py` uses `os.walk()` to recursively collect YAML files from the `source/` directory. However, this includes all subdirectories — specifically the `source/archive/` directory which contains outdated versions (e.g. 1.22) that should not be processed together with current active files.

**Current Behavior:**  
- Function processes 65 YAML files total  
- Includes archived files from `source/archive/` directory  
- Debug output confirms: `found 65 files of type yaml`  
- Archive files pass filename filters and get processed alongside current files  

**Expected Behavior:**  
- Function should exclude `archive/` directory from recursive walk  
- Should process only active YAML files directly under `source/` (~43 files)  
- Archive files should be completely ignored during processing  

**Evidence:**  
1. Debug output shows 65 files found vs expected ~43  
2. Archive directory contains files(~22) like `webapp-cards-1.22-en.yaml`  
3. Script currently processes both current (3.0) and archive (1.22) versions simultaneously  

**Impact:**  
- Version Conflicts: Old archived content mixed with current active content  
- Data Inconsistency: Generated documents may contain outdated information  
- User Confusion: Unexpected output when requesting specific versions  
- Build Issues: Potential conflicts between different version formats  

**Recommended Fix:**  
Modify `get_files_from_of_type()` to exclude the archive directory during recursion:

```python
for root, dirnames, filenames in os.walk(path):
    # Exclude archive directory from recursion
    dirnames[:] = [d for d in dirnames if d != 'archive']
    for filename in fnmatch.filter(filenames, "*." + str(ext)):
        files.append(os.path.join(root, filename))
```

sydseter:
We can live with this.

PR #2590, opened 2026-03-10
Fix: Exclude archive directory from YAML file collection

khushal-winner:
Closes - #2587 

- get_files_from_of_type() now excludes archive/ directory from os.walk()
- Reduces processed YAML files from 65 to 43 (excludes 22 archive files)
- Prevents version conflicts and mixed content in generated documents
- Fixes data integrity issue in build process

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -567,7 +567,7 @@ def get_files_from_of_type(path: str, ext: str) -> List[str]:
     files = []
     for root, dirnames, filenames in os.walk(path):
         # Exclude archive directory from recursion
-        dirnames[:] = [d for d in dirnames if d != 'archive']
+        dirnames[:] = [d for d in dirnames if d != "archive"]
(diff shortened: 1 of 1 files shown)


Suresh-Krishna-P:
Hello @khushal-winner, The fix for the `get_files_from_of_type()` function is a solid solution that effectively addresses the core issue. By adding a simple line to filter out the 'archive' directory during the `os.walk()` recursion, the code now properly excludes archived YAML files from being processed alongside current active files. This prevents the version conflicts and data inconsistency that were occurring when both 1.22 archived versions and 3.0 current versions were being mixed together. It's exactly the kind of targeted fix that solves the problem without introducing unnecessary complexity. The change reduces the processed files from 65 to 43, which aligns perfectly with the expected behavior of only processing active content while leaving the archived material untouched for reference purposes. Well Done.

PR #2593, opened 2026-03-10
Add health check endpoint for Fly.io deployment

khushal-winner:
- Create HealthController with database connectivity check
- Add /health route using API pipeline for efficiency
- Configure Fly.io HTTP health checks with proper timing
- Prevents traffic routing to instances before DB is ready

Fixes issue where Fly.io TCP port check marked instances as healthy before database migrations and connectivity were established.


sydseter:
Thank you for the update. This is actually very helpful.

PR #2598, opened 2026-03-11
fix: restrict Next Round button during voting phase until all votes are cast

### LINKED ISSUE #2592: Bug: Any player can skip voting phase by  clicking "Next Round" button
**Describe the bug**
During the voting phase (after all players have played their cards), the "Next Round" button has no restrictions and is fully enabled for all players. Any player can click it to skip voting entirely, preventing other players from scoring their votes.

**To Reproduce**
1. Go to copi.owasp.org and start a game with 2+ players
2. All players play their cards in the current round
3. During the voting phase, any player clicks "Next Round" button
4. Round advances immediately skipping all pending votes

**Expected behavior**
The "Next Round" button should only be clickable after all players have finished voting, or should be restricted to the game host only. Players should not be able to skip the voting phase.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: Latest

**Additional context**
Root cause identified in the code:
- In show.html.heex, the button is only disabled when round_open? is TRUE and can_continue_round? is FALSE
- During voting phase, round_open? is FALSE, so can_continue_round? always returns FALSE (it requires round_open? to be TRUE)
- This means the Next Round button is always fully enabled during voting with zero protection

Relevant files:
- copi.owasp.org/lib/copi_web/live/player_live/show.html.heex
- copi.owasp.org/lib/copi/cornucopia/game.ex
- copi.owasp.org/lib/copi_web/live/player_live/show.ex

- [x] Yes
- [ ] No

pranitaurlam:
## Summary

Fixes #2592 — Any player could skip the voting phase by clicking "Next Round" because the button was only disabled when `round_open? == true AND can_continue_round? == false`. During the voting phase (`round_open? == false`), the button had no restrictions.

- Added `all_votes_cast?/1` helper that checks every player has voted on every other player's card in the current round
- Added server-side guard in `handle_event("next_round", ...)`: rejects advancement when `round_closed?` but votes are incomplete, logging a warning
- Updated the "Next Round" / "Finish game" button in `show.html.heex` to also disable during the voting phase until all votes are cast

## Test plan

- [ ] `next_round is rejected when round is closed but not all votes are cast` — 2 players play cards, no votes cast → rounds_played stays at 0
- [ ] `next_round advances when round is closed and all votes are cast` — 2 players play cards, cross-vote → rounds_played increments to 1
- [ ] `all_votes_cast?/1 returns true when all players have voted on all cards`
- [ ] `all_votes_cast?/1 returns false when votes are missing`
- [ ] Existing tests continue to pass (1-player games satisfy the check trivially since 0 votes needed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/copi.owasp.org/lib/copi_web/live/player_live/show.ex
+++ b/copi.owasp.org/lib/copi_web/live/player_live/show.ex
@@ -81,10 +81,14 @@ defmodule CopiWeb.PlayerLive.Show do
         {:noreply, socket}
       end
     else
-      Copi.Cornucopia.update_game(game, %{rounds_played: game.rounds_played + 1, round_open: true})
+      if all_votes_cast?(game) do
+        Copi.Cornucopia.update_game(game, %{rounds_played: game.rounds_played + 1, round_open: true})
 
-      if last_round?(game) do
-        Copi.Cornucopia.update_game(game, %{finished_at: DateTime.truncate(DateTime.utc_now(), :second)} )
+        if last_round?(game) do
+          Copi.Cornucopia.update_game(game, %{finished_at: DateTime.truncate(DateTime.utc_now(), :second)} )
+        end
+      else
+        Logger.warning("next_round rejected: not all votes cast. game_id: #{game.id}")
       end
     end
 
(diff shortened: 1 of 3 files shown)


pranitaurlam:
Hi @sydseter! I've opened a PR to fix this issue: https://github.com/OWASP/cornucopia/pull/2598

Root cause confirmed: In show.html.heex, the "Next Round" button was only disabled when round_open? == true AND can_continue_round? == false. During the voting phase (round_open? == false), the button had no restrictions at all, allowing any player to skip voting.

Approach taken Option 1 (disable until all players have voted):

Added all_votes_cast?/1 helper that checks every player's card in the current round has received num_players - 1 votes (one from each other player)
Added a server-side guard in handle_event("next_round", ...) that rejects advancement when the round is closed but votes are incomplete
Updated the button's disabled condition in the template to also disable during the voting phase until all votes are cast
Added tests covering the new behaviour
Happy to switch to Option 2 (restrict to game host only) if that's preferred  just let me know!

Adarshkumar0509:
@pranitaurlam I really appreciate your hardwork but this issue was assigned to me.

sydseter:
The point with skip to next round is to allow players to skip to the next round without having cast their vote.

pranitaurlam:
@sydseter may i know what is the issue to close my pr

Adarshkumar0509:
@pranitaurlam The reason sydseter closed your PR is because the Next Round button allowing players to skip voting is intentional game design. sydseter mentioned The point with skip to next round is to allow players to skip to the next round without having cast their vote. So this is not a bug but a feature players are allowed to skip voting if they choose to. Thanks for contributing to our project.

PR #2613, opened 2026-03-12
Allow exclude some tags from validation for untranslation status

### LINKED ISSUE #2612: When release created it list some values as untranslated
I did take a look at release notes, to check what's missing in Ukrainian translation and notice that all "errors" comes from strings which shouldn't be translated

T02330, T02530 - Cornucopia (this is brand name)
T03130, T03150, T03170, T03190, T03240, T03260, T03350, T03420, T03470, T03490, T03540, T03580, T03710, T03730, T03750, T03770, T03772, T03774 - Application version (it's just numbers and obviously should not be translated)

So can these values to be excluded from the report, otherwise it would be hard to keep an eye on the translation quality.

kant2002:
Closes #2612

### DIFF
--- a/scripts/check_translations.py
+++ b/scripts/check_translations.py
@@ -19,8 +19,9 @@ from collections import defaultdict
 class TranslationChecker:
     """Check translations for missing, untranslated, or empty tags."""
 
-    def __init__(self, source_dir: Path):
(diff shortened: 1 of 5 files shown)


sydseter:
Thanks for your help!

PR #2614, opened 2026-03-12
Correct Norwegian translation

sydseter:
Fixes remaining issues with the Norwegian translation.


sydseter:
@copilot check if the Norwegian translation to source/webapp-cards-3.0-no_nb.yaml from source/webapp-cards-3.0-en.yaml is satisfactory.

PR #2616, opened 2026-03-12
Fix: remove hardcoded version in card route loader and use DeckServic…

### LINKED ISSUE #2603: Question: Hardcoded edition version in card route loader
While exploring the card routing logic I noticed that version resolution is handled differently in two similar routes.

In:

`src/routes/edition/[edition]/+page.server.ts`

the version is determined using:

```ts
const version = DeckService.getLatestVersion(edition);
```

However in:

`src/routes/edition/[edition]/[card]/+page.server.ts`

the version is hardcoded:

```ts
const version = edition == 'webapp'
  ? '2.2'
  : edition == 'mobileapp'
  ? '1.1'
  : '1.0';
```

I wanted to ask whether this difference is intentional (for example to lock the card route to specific versions), or if it would make sense to use `DeckService.getLatestVersion()` here as well for consistency.

Happy to submit a PR if aligning these would be appropriate.

kavya-seth-vns:
Fixes #2603

The card route loader previously used hardcoded version mappings:

* webapp → 2.2
* mobileapp → 1.1
* other → 1.0

This PR replaces the hardcoded logic with `DeckService.getLatestVersion(edition)`, aligning it with the implementation already used in `src/routes/edition/[edition]/+page.server.ts`.

This removes hardcoded version handling and ensures both routes resolve edition versions consistently.

### DIFF
--- a/scripts/capec_map_enricher.py
+++ b/scripts/capec_map_enricher.py
@@ -20,9 +20,7 @@ import yaml
 
 class EnricherVars:
     TEMPLATE_FILE_NAME: str = "EDITION-capec-VERSION.yaml"
-    DEFAULT_CAPEC_JSON_PATH = (
-        Path(__file__).parent / "../cornucopia.owasp.org/data/capec-3.9/3000.json"
-    )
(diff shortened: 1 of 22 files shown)


sydseter:
It's hardcoded for a reason. I will close this.

PR #2621, opened 2026-03-13
Fix crash on empty YAML files in convert.py

khushal-winner:
- Add null checks after yaml.safe_load() calls
- Prevent AttributeError when data is None
- Handle both empty files and explicit null content
- Protect build system from crashing on empty YAML files

Fixes issue where empty YAML files in source/ directory would cause: AttributeError: 'NoneType' object has no attribute 'keys'

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -636,8 +636,8 @@ def get_mapping_data_for_edition(
     with open(mappingfile, "r", encoding="utf-8") as f:
         try:
             data = yaml.safe_load(f)
-            # Handle empty/null YAML files
-            if data is None:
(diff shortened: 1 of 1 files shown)


sydseter:
@khushal-winner Try to focus on one pull-request at a time. Believe me, it’s not easier. Make sure you check the buil output when a build fails in the ci/cd pipeline.

khushal-winner:
> @khushal-winner Try to focus on one pull-request at a time. Believe me, it’s not easier. Make sure you check the buil output when a build fails in the ci/cd pipeline.

yess i will check before asking for a review , & will focus on only one PR  :)

PR #2707, opened 2026-03-20
fix: handle_event(next_round) missing early return causes double broadcast (#2418)

### LINKED ISSUE #2418: bug: handle_event("next_round") missing early return causes double broadcast and premature game state update
## Description

In `copi.owasp.org/lib/copi_web/live/player_live/show.ex`, the `handle_event("next_round")` function has a control-flow bug caused by missing early returns in Elixir.

## Root Cause

The code attempts to return early from a `cond`/`if` block, but **Elixir does not have early returns** — `if` is an expression whose value is discarded unless it is the last expression in the function. This means the final three lines always execute regardless of which branch was taken:

```elixir
def handle_event("next_round", _, socket) do
  game = socket.assigns.game

  if round_open?(game) do
    if Copi.Cornucopia.Game.can_continue_round?(game) do
      Copi.Cornucopia.update_game(game, %{round_open: false})
      Process.send_after(self(), :proceed_to_next_round, 100)
      {:noreply, assign(socket, :game, game)}  # ← VALUE DISCARDED, not an early return
    else
      {:noreply, socket}  # ← VALUE DISCARDED
    end
  else
    Copi.Cornucopia.update_game(game, %{rounds_played: game.rounds_played + 1, round_open: true})
    if last_round?(game) do
      Copi.Cornucopia.update_game(game, %{finished_at: ...})
    end
  end

  # ↓ These ALWAYS execute — even when the intent was to wait for the 100ms timer
  {:ok, updated_game} = Game.find(game.id)
  CopiWeb.Endpoint.broadcast(topic(updated_game.id), "game:updated", updated_game)
  {:noreply, assign(socket, :game, updated_game)}
end
```

## Impact

When `round_open?(game)` is **true** and `can_continue_round?` is **true**:

1. `round_open` is set to `false`
2. A 100ms timer is scheduled via `Process.send_after(self(), :proceed_to_next_round, 100)`
3. **Immediately** (before the timer), the code also calls `Game.find` + broadcasts the game state as-is (with `rounds_played` not yet incremented)
4. 100ms later, `:proceed_to_next_round` fires and broadcasts **again** with `rounds_played + 1`

This causes:
- **Double broadcast**: clients receive two rapid `game:updated` events, potentially causing UI flickering or state inconsistency
- **Premature broadcast**: the first broadcast sends `round_open: false` but with the old `rounds_played`, confusing all connected LiveView clients
- **Race condition**: the 100ms timer in `handle_info(:proceed_to_next_round)` runs the same update+broadcast logic, doubling the work

immortal71:
## Summary

Fixes #2418

In copi.owasp.org/lib/copi_web/live/player_live/show.ex, the handle_event(next_round) function had a control-flow bug: Elixir has no early returns, so the outer if/else block was not the last expression in the function. This caused the final three lines (Game.find, Endpoint.broadcast, {:noreply, ...}) to **always execute** unconditionally, regardless of which branch ran.

## Root Cause

`elixir
  if round_open?(game) do
    ..
    {:noreply, assign(socket, :game, game)}  # value discarded - NOT an early return
  else
    ..
  end

  These always ran -- even when waiting for the 100ms timer
  {:ok, updated_game} = Game.find(game.id)
  CopiWeb.Endpoint.broadcast(...)
  {:noreply, assign(socket, :game, updated_game)}
`

# Fix

Moved Game.find + Endpoint.broadcast + {:noreply, ...} into the else branch so the outer if/else is the last expression of the function. Each branch now returns the correct value


immortal71:
@sydseter  is this good to go ??

Issue #2232, opened 2026-02-14
Encrypt the game name and player name fields for copi.owasp.org


sydseter:
We should make sure the game name and player name fields for copi.owasp.org is encrypted. During migration, the data not encrypted should get encrypted by the application. The key need to be stored as an environment variable when deployed.

Klath123:
Hi @sydseter , I would like to work on this issue.

I can implement encryption and decryption for the game name and player name fields, ensure existing unencrypted data is securely migrated, and manage the encryption key through environment variables as suggested. I will also make sure the changes integrate cleanly with the existing application workflow.
Please let me know if I can proceed and if there are any preferred encryption approaches or libraries for this project.

sydseter:
I have assigned you.

aadilrizwan:
@sydseter can i work on this?

sydseter:
@InFiNiTy0639 Have a look at the issues that not is assigned to anyone.

Arunodoy18:
Hi 👋

I’d like to take this on.

Proposed approach:

- Implement field-level encryption for `game.name` and `player.name` using application-level encryption (not database-level).
- Store the encryption key securely via environment variable (e.g. COPI_ENCRYPTION_KEY).
- Introduce a migration strategy:
  - Keep existing plaintext data readable.
  - Encrypt legacy records during application boot or via a one-time migration task.
- Ensure decryption happens transparently at the schema level so existing queries continue to work.

I will make sure:
- No breaking changes to existing flows.
- Encryption key is never hardcoded.
- Backward compatibility is preserved during transition.

Please let me know if there are any preferred encryption libraries or patterns already used in the project before I proceed.

Adarshkumar0509:
hii @sydseter, i want to work on this issue Could you please assign this issue to me?

Adarshkumar0509:
@sydseter sir can i take a look at this issue?

sydseter:
@Adarshkumar0509 Have a go at it.

sydseter:
@Arunodoy18 I reassigned this issue to @Adarshkumar0509

Adarshkumar0509:
Hi @sydseter, i seen that the master currently has 54 pre-existing test failures related to a missing scores column on dealt_cards. my PR does not introduce any of these all 13 encryption specific tests pass cleanly and our changes only touch games.name and players.name.

PR #2743, opened 2026-03-27
feat(security): encrypt game and player names with env-based key

Adarshkumar0509:
Implements field-level encryption for game.name and player.name as requested in #2232.

Uses Cloak and Cloak.Ecto for AES-256-GCM authenticated encryption. The vault
is configured via the COPI_ENCRYPTION_KEY environment variable in production.
A deterministic test key is configured in test.exs so CI runs without any
additional setup.

The database migration converts the name columns from varchar to binary.
Existing plaintext data is preserved during migration and remains readable
until re-encrypted on next write.

Tested locally with PostgreSQL — 13 tests passing.

Resolved or fixed issue: #2232

###Ai disclosure 
Used AI to assist with implementation. The logic, integration with the existing schema and application layer, and all verification were done manually including running migrations and tests locally. No AI-generated noise or unnecessary comments are present in the code. All code are manually written.


Adarshkumar0509:
@sydseter please check this out ,Happy to make any changes you suggest.

sydseter:
@Adarshkumar0509 That won't work, we will need to migrate the data and encrypt it in order to be able to load existing player names. Otherwise you get a mismatch between the strings that is saved in the database and the new binary type.

Adarshkumar0509:
Ok sir I will be working on this

PR #2827, opened 2026-04-16
fix: replace structural <span> wrappers with semantic headings (<h1>, <h2>)

### LINKED ISSUE #2826: Fix main headings being rendered as `<span>` tags on the homepage
### Description
While reviewing the landing page in `src/routes/+page.svelte`, I noticed that the main titles of the page are currently built using <span> tags instead of standard heading tags.
For example, on line 35 we have `<span id="top">{$t('home.h1.1')}</span>`, and similar setups are used on lines 45 and 61. Even though the translation keys (h1.1, etc.) suggest these should be headers, they are just styled as large text rather than being real structural headings in the code.

### Impact
- Accessibility: Screen readers rely on real headings (`<h1>`, `<h2>`) so visually impaired users can navigate the page. Using `<span>` tags breaks this navigation completely.
- SEO: Search engines look for `<h1>` tags to figure out what a website is about. Missing them can hurt the site's search ranking.
- Future Styling: Making all `<span>` tags large and bold by default (as done on line 69) makes it hard to use normal, small text spans anywhere else on the page later.

### Proposed Solution
I'd love to submit a quick PR to fix this! The fix is very simple:

- Change the first title wrapper (Line 35) to an `<h1>` tag.
- Change the secondary title wrappers (Lines 45 and 61) to `<h2>` tags.
- Update the CSS at line 69 so it only applies the large, bold styling to h1, h2 instead of every span on the page.

Mysterio-17:
### Description
This PR resolves an accessibility (a11y) and SEO issue on the landing page where the primary page titles were being rendered using un-classed <span> tags rather than proper HTML document headings.
Previously, home.h1.1, home.h1.2, and home.h1.3 were wrapped in <span> tags, which broke screen-reader navigation and search-crawler context. To make them visually appear as headers, a global styling block was originally applied to all span components, which made future inline styling brittle.

### Solution

- src/routes/+page.svelte: Replaced the top-most main component wrapper (home.h1.1) with a single `<h1>` tag to establish the root document hierarchy.
- src/routes/+page.svelte: Converted the subsequent structural wrappers (home.h1.2 and home.h1.3) to `<h2>` tags.
- CSS Formatting: Updated the Svelte style block to directly target the new h1, h2 headings instead of globally hijacking the span element.

### AI Usage
AI was used to help format this PR description. The code analysis, HTML/CSS changes, and solution logic are entirely my own work and free of AI-generated noise and unnecessary comments.

### DIFF
--- a/copi.owasp.org/lib/copi/encrypted/binary.ex
+++ b/copi.owasp.org/lib/copi/encrypted/binary.ex
@@ -21,10 +21,7 @@ defmodule Copi.Encrypted.Binary do
   @impl Ecto.Type
   def dump(nil), do: {:ok, nil}
   def dump(value) when is_binary(value) do
-    case encrypt(value) do
-      {:ok, blob} -> {:ok, blob}
-      {:error, reason} -> raise "Copi.Encrypted.Binary dump/1 failed: #{reason}"
-    end
+    encrypt(value)
   end
   def dump(_), do: :error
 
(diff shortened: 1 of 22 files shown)


Mysterio-17:
Hey @sydseter , have a look at this whenevr you have time.

sydseter (approved):

sydseter:
@Mysterio-17 Thank you for your time!

PR #3236, opened 2026-07-17
fix: switch rate limiter to Fly-Client-IP for fly deployments (closes…

### LINKED ISSUE #3227: Swith the rate limiter from using X-Forwarded-For to be using Fly-Client-IP for a fly setup
What can go Wrong?

The current rate-limiting design is easy to evade in common proxy or multi-instance deployments. The rate limiting trusts the left-most X-Forwarded-For value and explicitly skips connection limiting when only remote_ip is available. The counters are kept in the in-memory GenServer. That means spoofed forwarded headers, multiple app nodes, or proxy/header misconfiguration can let an attacker bypass the main abuse-control mechanism. We are more than happy to get suggestions for how to improve the rate-limiting. Currently we are not seeing a lot of issues concerning the misuse of our services in production.

The Fly.io reverse proxy does not strip or rewrite untrusted X-Forwarded-For headers before traffic reaches Phoenix. It is therefor still possible to circumvent the rate-limiter. 

What are we going to do about it?

Swith the rate limiter from using X-Forwarded-For to be using Fly-Client-IP for a fly setup. Make the change configurable and use X-Forwarded-For by default. We don't know whether this really will work, but we can test it out.

Adarshkumar0509:
closes #3227

### Description

- Fly.io doesn't strip X-Forwarded-For headers so attackers can spoof them and bypass the rate limiter. switched to Fly-Client-IP which Fly sets itself and can't be spoofed.

- configurable via USE_FLY_CLIENT_IP=true, defaults to X-Forwarded-For so nothing breaks on non-Fly setups.

### AI Tool Disclosure

- [X] My contribution does not include any AI-generated content
- [ ] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie etc.]`
    - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro etc.]`
    - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

### Affirmation

- [X] My code follows the [CONTRIBUTING.md](https://github.com/owasp/cornucopia/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/copi.owasp.org/lib/copi/ip_helper.ex
+++ b/copi.owasp.org/lib/copi/ip_helper.ex
@@ -100,18 +100,27 @@ defmodule Copi.IPHelper do
   # Private helpers
 
   defp get_forwarded_ip(conn) do
-    # Get X-Forwarded-For header (rightmost IP is most recent proxy, leftmost is original client)
-    case Plug.Conn.get_req_header(conn, "x-forwarded-for") do
-      [] -> nil
-      [forwarded | _] ->
-        # Take the first (leftmost) IP from the comma-separated list
-        forwarded
-        |> String.split(",")
-        |> List.first()
-        |> String.trim()
-        |> parse_ip_string()
-      _ -> nil
+    if fly_client_ip_enabled?() do
+      case Plug.Conn.get_req_header(conn, "fly-client-ip") do
+        [ip | _] -> parse_ip_string(String.trim(ip))
+        [] -> nil
+      end
+    else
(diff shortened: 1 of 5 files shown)


Adarshkumar0509:
hii @sydseter whenever you have time please take a look at this.

sydseter (approved):

PR #3298, opened 2026-07-28
fix(scripts): return None from get_docx_document on missing template

### LINKED ISSUE #2490: Bug: get_docx_document() returns truthy blank Document on missing template, silently writing empty output file
## Describe the bug

In `scripts/convert.py`, the `get_docx_document()` function is supposed to open and return a `.docx` template file. When the file does not exist, it logs an error but **returns a blank `docx.Document()` instead of `None`**. Since a blank `docx.Document()` is always truthy in Python, the caller's `if doc:` check always passes — causing the script to silently replace text in and save an empty document to the output path, producing a corrupted zero-content output file with no further error.

## Affected file

`scripts/convert.py` — `get_docx_document()` and its call site in `create_edition_from_template()`

## Code snippet

```python
def get_docx_document(docx_file: str) -> Any:
    import docx
    if os.path.isfile(docx_file):
        return docx.Document(docx_file)
    else:
        logging.error("Could not find file at: %s", str(docx_file))
        return docx.Document()  # <-- blank Document, always truthy!

# Call site:
doc = get_docx_document(template_doc)
if doc:  # always True — even for a blank Document returned on error
    doc = replace_docx_inline_text(doc, language_dict)
    doc.save(output_file)  # saves an empty file with no content
```

## Expected behavior

When the template file does not exist:
- `get_docx_document()` should return `None`
- The caller should detect `None` and skip the output file creation, or raise a clear error

## Proposed fix

```python
def get_docx_document(docx_file: str) -> Any:
    import docx
    if os.path.isfile(docx_file):
        return docx.Document(docx_file)
    else:
        logging.error("Could not find file at: %s", str(docx_file))
        return None  # caller already checks `if doc:`
```

## Steps to reproduce

1. Specify a non-existent or missing template `.docx` file path
2. Run `python scripts/convert.py -lt cards -l en -v 2.2`
3. No error is raised; an empty `.docx` file is written to the output folder

## Additional context

The docstring does not indicate that a blank fallback document will be returned on failure, making this behavior surprising and hard to diagnose. Returning `None` is cleaner, consistent with the existing `if doc:` guard in the caller, and avoids silently writing corrupt output files.

prajakta128:
Resolves #2490

## Summary
`get_docx_document()` was returning a blank `docx.Document()` object when the template file could not be found. Because a blank `docx.Document()` is always truthy in Python, the caller's `if doc:` check in `create_edition_from_template()` always passed — so a missing template silently produced an empty, corrupt `.docx` output file instead of failing with a clear error.

## Fix
- `get_docx_document()` now returns `None` instead of a blank document  when the file is missing. Return type updated from `Any` to  `Optional[Any]` to reflect this.
- `create_edition_from_template()` now checks `if doc is not None:`  instead of `if doc:`, and adds an `else` branch that logs a clear error (`Cannot create output file: template document not found at %s`) and returns early — so no empty file gets written.

## Scope
Only `scripts/convert.py` and `tests/scripts/convert_utest.py` are touched. No unrelated files, formatting, or refactors are included.

## Testing
Ran locally before opening this PR:
- `pytest tests/scripts/convert_utest.py` — 149/149 pass on Linux
- `mypy --namespace-packages --strict ./scripts/` — clean
- `flake8 --max-line-length=120 --max-complexity=10` — clean
- `black --check` — clean

One note: on Windows, `test_get_template_abs_path` fails locally due to path-separator handling unrelated to this change (fails identically on `master` without this PR applied) — flagging it here for visibility, happy to open a separate issue for it if useful.

---
This is my first PR to Cornucopia — feedback very welcome, and happy to adjust anything that doesn't fit the project's conventions.

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -11,7 +11,7 @@ import yaml
 import zipfile
 import xml.etree.ElementTree as ElTree
 from defusedxml import ElementTree as DefusedElTree
-from typing import Any, Dict, List, Tuple, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 from operator import itemgetter
 from itertools import groupby
 from pathlib import Path
@@ -381,9 +381,12 @@ def create_edition_from_template(
         if file_extension == ".docx":
             # Get the input (template) document
             doc = get_docx_document(template_doc)
-            if doc:
+            if doc is not None:
                 doc = replace_docx_inline_text(doc, language_dict)
                 doc.save(output_file)
+            else:
+                logging.error("Cannot create output file: template document not found at %s", template_doc)
+                return
         else:
             save_odt_file(template_doc, language_dict, output_file)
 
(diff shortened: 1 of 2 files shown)


sydseter (approved):

sydseter:
@prajakta128  Thank you. all commits must have a verified signature.

PR #3301, opened 2026-07-29
fix(cornucopia.owasp.org): enforce pnpm usage and prevent npm installs

### LINKED ISSUE #2623: Prevent mixing npm and pnpm in the project
### Problem

The project currently uses **pnpm** for dependency management, but it is still possible to run commands using **npm**.

Because of this, contributors may accidentally run:

```bash
npm install
```

instead of:

```bash
pnpm install
```

This can introduce conflicts in the repository.

---

### Conflicts that can happen

When **npm and pnpm are used side by side**, several problems may occur:

- Running `npm install` creates a **package-lock.json** file, while the project already uses **pnpm-lock.yaml**.
- npm installs dependencies using a **flat `node_modules` structure**, while pnpm uses a stricter and different layout.
- pnpm-specific configuration in `package.json` may be ignored when npm is used.
- The dependency tree may become inconsistent between contributors' environments.
- Builds or scripts may behave differently depending on which package manager was used.

These issues make debugging harder and reduce reproducibility for contributors and CI.

---

### Proposed solution

To prevent these issues, the project could enforce pnpm usage by:

1. Adding the following field to `package.json`:

```json
"packageManager": "pnpm@10.3.0"
```

2. Optionally adding a preinstall check to prevent npm or yarn installs:

```json
"preinstall": "npx only-allow pnpm"
```

3. Updating documentation to consistently use:

```bash
pnpm install
pnpm run <script>
```

---

### Question

If this sounds like a useful improvement, **may I work on implementing this change?**

This would be my first contribution to the project, and I would be happy to help improve the setup for contributors.

prajakta128:
Resolves #2623

## Summary
The project uses pnpm for dependency management, but nothing prevented contributors from accidentally running `npm install`, which creates a conflicting `package-lock.json` alongside the existing `pnpm-lock.yaml` and can produce inconsistent dependency trees between contributors.

## Changes
Scoped to `cornucopia.owasp.org/` only:
- Added `"packageManager": "pnpm@<version>"` to `package.json`
- Added a `preinstall` script using `only-allow pnpm` to block npm/yarn installs
- Updated README references from `npm install`/`npm run` to the `pnpm` equivalents

## Testing
- `npm install` inside `cornucopia.owasp.org/` now fails as expected via the preinstall guard
- `pnpm install` still works normally

## Notes
A prior PR (#2624) attempted this fix but was closed for inactivity before two small requested changes were addressed (an unrelated `.pre-commit-config.yaml` version bump, and a duplicate README headline). This PR keeps strictly to the three changes above and avoids both of those.

### DIFF
--- a/cornucopia.owasp.org/package.json
+++ b/cornucopia.owasp.org/package.json
@@ -2,7 +2,9 @@
 	"name": "cornucopia-website",
 	"version": "0.0.1",
 	"private": true,
+	"packageManager": "pnpm@10.0.0",
 	"scripts": {
+		"preinstall": "npx only-allow pnpm",
 		"dev": "vite dev",
 		"prebuild": "echo I run before the build script",
 		"productionbuild": "vite build && node ./script/headers.js && npx svelte-sitemap --domain https://cornucopia.owasp.org --ignore 404 --ignore cards/COM* --ignore cards/DVE* --ignore cards/AC* --ignore cards/CO*",


sydseter:
@prajakta128 Thank you. Please ensure commits have a verified signature.

sydseter (approved):

sydseter:
@prajakta128 Thank you for your efforts. If you’d like, please add your name to the acknowledgment list. https://github.com/OWASP/cornucopia/blob/master/cornucopia.owasp.org/data/website/pages/tribute/en/index.md#volunteers

Issue #2559, opened 2026-03-07
🚨Security Issue: play_card API Vulnerable to DoS Attack - 29% Success Rate Confirmed


khushal-winner:
**Describe the bug** (watch the full video)
The API endpoint `PUT /api/games/:game_id/players/:player_id/card` (play_card action) has no rate limiting protection. Testing shows 100 simultaneous requests are all processed, with 29 succeeding and 71 returning 406 errors. This creates a resource exhaustion vulnerability where attackers can flood the database with concurrent operations and disrupt game state integrity.

**Expected behavior**
API should implement rate limiting to reject excessive requests with HTTP 429 status after a reasonable threshold (e.g., 10 requests per minute per IP). Only the first valid request should succeed; subsequent duplicate requests should be blocked before database operations.

**Video Demo**

https://github.com/user-attachments/assets/4181b3c5-12d0-4a31-bee9-84f390b78e32

**Desktop (please complete the following information):**
OS: Any (tested on Windows)
Browser: Chrome, Firefox, Safari
Version: Latest

**Additional context**
- Vulnerability confirmed via browser console attack script
- Each request triggers database queries and game state updates
- Race conditions allow multiple successful card plays (29% success rate)
- WebSocket broadcasts triggered for each successful play
- Production server vulnerable to DoS attacks
- Fix: Add RateLimiterPlug to :api pipeline in router.ex

**Worst-Case Scenario**
- Complete Service Outage
- Attacker launches 10,000 concurrent requests
- Database connection pool exhausted
- Server crashes under load
- Service requires manual restart

**Script Used**
```
// More sophisticated attack with timing
const attack = {
  gameId: "01KK4ANZFV6XMR14VX7R1X6T55",
  playerId: "01KK4APC64N6Z5B346S960XTQJ",
  cardId: "47369",
  
  async sendRequest(requestNum) {
    try {
      const response = await fetch(`https://copi.owasp.org/api/games/${this.gameId}/players/${this.playerId}/card`, {
        method: 'PUT',
        headers: { 'Content-Type': 'application/json' },
        body: JSON.stringify({ dealt_card_id: this.cardId })
      });
      
      console.log(`Request ${requestNum}: ${response.status} ${response.statusText}`);
      return response.status;
    } catch (error) {
      console.error(`Request ${requestNum} failed:`, error);
      return 'ERROR';
    }
  },
  
  async launchAttack(count = 100) {
    console.log(`🚀 Starting attack with ${count} requests...`);
    
[...]

PR #3310, opened 2026-08-01
fix: rate-limit API endpoints and make card plays atomic

khushal-winner:
Fixes OWASP Cornucopia security issue on `PUT /api/games/:game_id/players/:player_id/card` (play_card): no rate limiting and a TOCTOU race allowing duplicate plays under concurrency.

Fixes - #2559

## Changes

- `copi.owasp.org/lib/copi/rate_limiter.ex` - add `:api` bucket (60 req / 60 s) and allow the action
- `copi.owasp.org/lib/copi_web/plugs/rate_limiter_plug.ex` - accept and forward the `:api` action
- `copi.owasp.org/lib/copi_web/router.ex` - wire `RateLimiterPlug action: :api` into `:browser_api`; remove the dead `:api` pipeline (previously router.ex:16-18) that never had the plug
- `copi.owasp.org/lib/copi_web/controllers/api_controller.ex` - atomic one-card-per-player-per-round play inside a transaction: player row locked with `SELECT ... FOR UPDATE`, then a conditional `UPDATE ... WHERE NOT EXISTS` on any other card already played by that player in the current round; `409` on already-played / player-missing, `503` on transaction failure
- `copi.owasp.org/fly.toml` - `RATE_LIMIT_API_LIMIT = 60`, `RATE_LIMIT_API_WINDOW = 60` in production env
- `copi.owasp.org/SECURITY.md` - document the API rate-limit bucket and env vars
- `copi.owasp.org/test/copi/rate_limiter_test.exs` - `:api` limiter tests
- `copi.owasp.org/test/copi_web/plugs/rate_limiter_plug_test.exs` - plug action / 429 tests
- `copi.owasp.org/test/copi_web/controllers/api_controller_test.exs` - 409, 10-way same-card concurrent, 2-way different-cards concurrent (exactly one 200 + one 409), transaction-failure 503, player-disappears 409, and an end-to-end 429 through the router when the `:api` bucket is exhausted

## Review feedback - addressed

Addressing @sydseter's review on #2559:

1. **router.ex:16-18 defined the API pipeline without RateLimiterPlug.** That pipeline was dead code (no scope ever piped through it). It has been removed. The scope that actually serves `/api` (`:browser_api`) pipes every request through `RateLimiterPlug, action: :api`, so `PUT /api/games/:id/players/:id/card` (and the other `/api` routes) are now rate limited.

2. **`play_card` performed a read/check followed by an unconditional update; concurrent requests could both succeed.** The update is now atomic and conditional inside a transaction: the player row is locked with `SELECT ... FOR UPDATE` to serialize plays by the same player, and the card update only applies when no other card by that player has been played in the current round (`UPDATE ... WHERE NOT EXISTS`). A losing request gets `409 Conflict` ("Card was already played by another request"). Exactly one play and one WebSocket broadcast per round.

3. **The existing limiter only limited browser connections.** The `:api` action bucket is enforced independently of the connection bucket and is configurable via `RATE_LIMIT_API_LIMIT` / `RATE_LIMIT_API_WINDOW`, defaulting to the suggested 60 requests per minute per IP.

4. **Endpoint-appropriate throttling plus an atomic/conditional database update.** Both are implemented: a general API throttle of 60 req/min per IP, and the atomic conditional play, each covered by tests (including an end-to-end 429 through the router).

Full suite: 485 tests, 0 failures, 95.3% coverage (CI threshold 95%).


sydseter:
@khushal-winner The solution works as intended, and the code looks ok. Thank you for your effort. That said, you have some merge conflicts that need to be resolved.

khushal-winner:
> @khushal-winner The solution works as intended, and the code looks ok. Thank you for your effort. That said, you have some merge conflicts that need to be resolved.

@sydseter thnx for the review, i will resolve the merge conflicts by tomorrow

sydseter (approved):

Issue #1322, opened 2025-06-02
Adding a endpoint for each eop card and to the card browser


sydseter:
Similarly as with cornucopia we should add an endpoint for each EoP card and have a browsable eop deck. A feature like that has been asked for through OWASP Threat Dragon in order to integrate threat modeling card games into Threat Dragon. This will make it easier to increase adoption when doing threat modeling using card games. 

Each of the cards in the EoP game should be browsable from here: https://cornucopia.owasp.org/cards
and from here: http://cornucopia.owasp.org/card/eop 

The code that need to be changed can be found here:

- https://github.com/OWASP/cornucopia/tree/master/cornucopia.owasp.org/src/routes/cards
- https://github.com/OWASP/cornucopia/tree/master/cornucopia.owasp.org/src/routes/card/%5Bedition%5D

The source for the cards (description and card values) are available here: https://github.com/OWASP/cornucopia/blob/master/source/eop-cards-5.0-en.yaml 

Once it is done, it should work like for the website edition:

https://cornucopia.owasp.org/cards/AZ2

You may need to add some dummy yaml pages under a folder called *eop-cards-1.0-en* here: https://github.com/OWASP/cornucopia/tree/master/cornucopia.owasp.org/data/cards

The YAML does not need to be filled out.

The graphics can be found here: https://github.com/adamshostack/eop/tree/master
We prefer the use of svg to minimize the site load, if possible.

immortal71:
@sydseter  Can you assign this to me ?

sydseter:
@immortal71 This one is meant for GSOC

immortal71:
@sydseter  Thanks for the clarification !!
I’m  surprise I had no idea cornucopia is participating in GSoC but I am happy,I am wanted work on project  with Cornucopia so it is good orportunity for me too!! 
                                                   I have 1 pr merge and 2-3 pr in progress (waiting for your suggestion as I have updated them as per u suggested ) ,I have no clue what to do for prep ?
                                                                                      Would you prefer that this work is kept for the selected GSoC contributor and that I focus on other issues in the meantime or  should I start working on a small proof-of-concept for the EoP card endpoints (based on the existing cards routes and YAML).
                                                                                                      Can you tell me what will be better ?

aadilrizwan:
[...]

PR #3331, opened 2026-08-08
Eop card browser/refactor mentor followups

### LINKED ISSUE #3403: Automating importing new decks updating existing decks for the Cornucopia Coupier at https://cornucopia.owasp.org/cards
The work focuses on turning the Cornucopia card pipeline into a generic, data-driven system so new decks and updated deck versions can be added with minimal manual effort. The goal is to let contributors drop in the required YAML, card content, styling, and image assets, and have the Svelte site automatically discover, load, and present the deck as a browsable experience at https://cornucopia.owasp.org/cards.

Today, parts of the loading pipeline are already convention-based, especially for mappings, styling, and card-image YAML files. The remaining effort is to remove the last hard-coded edition registrations, static card lists, and per-deck UI assumptions so deck discovery, routing, rendering, and browsing all derive directly from deck metadata. Once that is done, importing a new deck or updating an existing one becomes a repeatable content operation rather than an application code change.

Deck discovery should be configurable for convenience and security. Instead of always scanning all matching files under source, discovery should support a controlled configuration layer that defines which directories, file patterns, editions, versions, and asset roots are allowed. This makes it easier to work with draft decks, test decks, and partial imports during development, while also reducing the risk of unintentionally exposing unreviewed or malformed content in production. A configurable allowlist or manifest-based discovery model would give maintainers explicit control over what the site may load, without giving up the long-term goal of automatic deck onboarding.

1. Deck discovery must stop being hard-coded in cornucopia.owasp.org/src/lib/services/deckService.ts. Today the app only knows decks from the static latests and decks arrays. 
A new YAML in source is invisible until this registry is updated. 

2. Suit discovery must stop being hard-coded in cornucopia.owasp.org/src/domain/suit/suitController.ts.
That file keeps its own fixed deck/version list and language map, separate from DeckService, so even if DeckService were made dynamic this controller would still block the new deck.

3. The card aggregation routes still enumerate specific editions in cornucopia.owasp.org/src/routes/cards/+page.server.ts and cornucopia.owasp.org/src/routes/cards/[card]/+page.server.ts.
Those pages manually merge mobileapp, webapp, companion, and eop. They need to iterate over discovered editions instead.

4. The edition browser UI is hard-coded for four decks in cornucopia.owasp.org/src/routes/edition/[edition]/+page.svelte.
It has fixed deck buttons, fixed per-deck intro text, and fixed default preview cards (PC2, VE2, AAI2, SP2). For generic onboarding, that page needs deck metadata from YAML, not per-edition branches in Svelte. Unknown editions and cards should still result in a proper 404. 

5. Card detail rendering still has edition-specific component registration in cornucopia.owasp.org/src/lib/components/cardFound.svelte and cornucopia.owasp.org/src/lib/components/cardPreview.svelte.
cardFound chooses taxonomy components from a fixed map, and cardPreview only registers an edition-specific mapping component for eop. If a new deck needs taxonomy/mapping rendering, there is no generic fallback config yet.

6. Unknown card IDs are currently rejected by a static allowlist in cornucopia.owasp.org/src/domain/card/cardIds.ts, and that gate is used by cornucopia.owasp.org/src/routes/cards/[card]/+page.svelte plus the edition card pages.
That has to be removed in favor of “card exists in loaded deck data” else 404.

7. Suit and card ordering are not generic in cornucopia.owasp.org/src/domain/suit/order.ts and cornucopia.owasp.org/src/domain/card/order.ts.
A new deck with new suit names or new card IDs will fall outside those maps. Ordering should come from YAML order, or from explicit metadata in the yaml configuration.

8. Source YAML plus images is still not enough by itself, because DeckService also expects scaffolded content folders under cornucopia.owasp.org/data/cards.
See cornucopia.owasp.org/src/lib/services/deckService.ts: it loads the deck from source/...cards...yaml, but then reads technical-note.md and explanation.md from data/cards/<edition>-cards-<version>-<lang>. The scaffolding step is still part of the runtime content model which means that it should be part of the build. If scaffolding already exist for all ready decks, than the scaffolding should do nothing. 

9. Image serving is not actually “drop images into source and they work”.
The image YAML uses literal public URLs like the ones in source/eop-card-images-5.0.yaml, and cornucopia.owasp.org/src/lib/components/cardPreview.svelte just uses that URL directly. There is no source-folder image discovery or asset copy pipeline here. The images still need to exist at a served web path, typically under static. Ensure image assets and yaml assets are bundled together and picked up by the scaffolding to ensure they become available later when the the card browser is loaded.

10. Deck naming and human-readable labels are still hard-coded in cornucopia.owasp.org/src/lib/services/deckServiceConsts.ts and in the page title helper inside cornucopia.owasp.org/src/routes/cards/[card]/+page.svelte.
A generic system needs edition display name, description, maybe preview card, and whether it is “external” to come from metadata.

ayman-art:
### Description

Applying mentor follow-ups across the EoP Card Browser project PRs.

**Done:**
- Added `decks.yaml` as the single registry of published editions/versions, with `DeckConfigService` to read it. This replaces hardcoded per-edition constants and branches across `DeckService`, `SuitController`, `CreController`, the CRE API, and the `/cards` + `/edition/[edition]` routes.
- `DeckService.getCards()` used instead of manually merging editions one by one.
- `deckService.ts` now discovers languages by scanning `source/` against what's declared in `decks.yaml`, instead of hardcoded `latests`/`decks` arrays.
- `domain/card/order.ts` and `domain/suit/order.ts` deleted because suit/card ordering now comes from source-file insertion order instead of separate maintained maps.
- Runtime validation for `decks.yaml` (`collectDeckConfigErrors`/`validateDecksYaml`): every required field is checked per deck, all problems across the file are collected and reported together, so a malformed entry fails loudly instead of silently.
- Fixed two bugs introduced while building this: eop's card-browser tooltip showed duplicated text ("Elevation of Privilege Elevation of Privilege version"), and the deck list on `/cards` could go stale after navigating because it wasn't set up to react to page updates.
- Documented the `decks.yaml` deck-registration workflow in `README.md`.
- Merged `master`, resolving conflicts across the four card-detail routes (config-driven ASVS/CAPEC resolution vs. hardcoded checks; adopted master's new `<Metadata>` component for page SEO tags).

**Current tasks:**
- Replace the hardcoded per-edition taxonomy/mapping component maps in `cardFound.svelte` / `cardPreview.svelte` with a generic, config-driven fallback.
- Add build-step scaffolding to generate `data/cards/{edition}-cards-{version}-{lang}/` content folders automatically.

Resolved or fixed issue:  closes #3403 & related to #1322
### AI Tool Disclosure

- [ ] My contribution does not include any AI-generated content
- [x] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `Claude Code`
    - LLMs and versions: `Claude Sonnet 5`
    - Prompts: [Refactor deck registration off hardcoded constants onto decks.yaml, add runtime validation for it;]
### Affirmation

- [x] My code follows the [CONTRIBUTING.md](https://github.com/owasp/cornucopia/blob/master/CONTRIBUTING.md) guidelines


ayman-art:
I want to note that for refactoring:

```
decks: decks.set(lang, new Map([...mobileCards, ...webappCards, ...companionCards, ...eopCards]))
```

I used `getCards()` existing in deckService, which loops over all editions. So it also adds `dbd`, which wasn't existing in the old map.

ayman-art:
@rewtd, @sydseter Question regarding cards/order.ts.. Before writing a script for this, do we really need controlling cards order **per suit**? If not, I can see the order is the same 2,3,4,5,6,7,8,9,X,J,Q,K,A for all suits. so we can use one shared ranking.

sydseter:
Yes, sure

ayman-art:
> Yes, sure

just confirming, you're good with one shared rank map, right?

cornucopia.owasp.org/src/lib/services/deckServiceConsts.ts
```diff
@@ -19,3 +19,13 @@ export const EDITION_NAMES: Record<string, string> = {
     [VERSION_DBD]: "Cornucopia",
     [VERSION_EOP]: "Elevation of Privilege"
 };
+
+/**
+ * Full edition name, used in card detail page titles and social meta tags in <svelte:head>.
+ */
+export const EDITION_FULL_NAMES: Record<string, string> = {
```
sydseter, line 26: This can be moved to yaml

sydseter (approved):

PR #2289, opened 2026-02-19
Fix race condition in voting system with unique constraint and safe upsert

immortal71:
## Summary

This PR fixes a critical race condition in the voting system (#2288) that could allow players to cast duplicate votes on the same card, corrupting game outcomes and data integrity.

## Problem

The voting system had two critical issues:

1. **Missing Database Constraint**: The `votes` table lacked a unique constraint on `(player_id, dealt_card_id)`, unlike the newer `continue_votes` table which properly includes such a constraint.

2. **Race Condition in Application Logic**: Both `toggle_vote` and `toggle_continue_vote` handlers used unsafe check-then-act patterns without proper transaction isolation, allowing concurrent requests to both pass the check and insert duplicate votes.

### Database Migration

- Added migration `20260218120000_add_unique_constraint_to_votes.exs`
- Cleans up any existing duplicate votes
- Creates unique index on `votes(player_id, dealt_card_id)`
- Ensures data integrity at the database level

### Application Code Refactoring

**lib/copi_web/live/player_live/show.ex**:

- Added `Vote` and `ContinueVote` aliases
- Added `Logger` requirement
- Refactored `toggle_continue_vote` to:
  - Use `Repo.get_by` for atomic vote checking
  - Use `on_conflict: :nothing` with `conflict_target` for safe inserts
  - Handle errors gracefully without crashing
  - Replace `IO.puts` debug statements with `Logger.debug/warning`

- Refactored `toggle_vote` to:
  - Use `Repo.get_by` instead of preloading and searching in memory
  - Use `on_conflict: :nothing` with `conflict_target` for safe inserts
  - Handle errors gracefully
  - Replace `IO.puts` debug statements with `Logger.debug/warning`

## Testing Strategy

The unique constraint ensures that:
1. Concurrent insert attempts will result in only one successful insert
2. The `on_conflict: :nothing` parameter prevents crashes on constraint violations
3. The application gracefully handles race conditions without user-visible errors

PR #3332, opened 2026-08-09
fix(copi): prevent duplicate votes via unique constraint and safe upsert

### LINKED ISSUE #2288: Race condition in voting system allows duplicate votes and data corruption
## Problem

The voting system in `copi.owasp.org` has a critical race condition that can lead to duplicate votes and data integrity issues. Both card voting and continue voting use a check-then-act pattern without proper database constraints or transaction isolation.

### 1. Missing Database Constraint on Card Votes

The `votes` table lacks a unique constraint on `(player_id, dealt_card_id)`:

```elixir
# priv/repo/migrations/20210816141921_create_votes.exs
create table(:votes) do
  add :player_id, references(:players, type: :uuid, on_delete: :nothing)
  add :dealt_card_id, references(:dealt_cards, on_delete: :nothing)
  timestamps()
end

create index(:votes, [:player_id])
create index(:votes, [:dealt_card_id])
# MISSING: unique constraint on [:player_id, :dealt_card_id]
```

### 2. Race Condition in Vote Toggle Logic

Both `toggle_vote` and `toggle_continue_vote` handlers in `lib/copi_web/live/player_live/show.ex` use unsafe check-then-act patterns:

**Card Voting (lines 125-145)**:
```elixir
vote = get_vote(dealt_card, player)

if vote do
  # Time window for race condition here
  Copi.Repo.delete!(vote)
else
  # Multiple concurrent requests can both pass this check
  case Copi.Repo.insert(%Copi.Cornucopia.Vote{
    dealt_card_id: String.to_integer(dealt_card_id), 
    player_id: player.id
  }) do
    {:ok, _vote} -> # ...
    {:error, _changeset} -> # ...
  end
end
```

**Continue Voting (lines 102-116)** has the same pattern.

## Testing

Add concurrent voting test:

```elixir
test "prevents duplicate votes from concurrent requests", %{game: game, player: player} do
  dealt_card = List.first(game.dealt_cards)
  
  # Simulate concurrent vote requests
  tasks = for _ <- 1..5 do
    Task.async(fn ->
      Repo.insert(%Vote{player_id: player.id, dealt_card_id: dealt_card.id})
    end)
  end
  
  results = Enum.map(tasks, &Task.await/1)
  successes = Enum.count(results, fn 
    {:ok, _} -> true
    _ -> false
  end)
  
  assert successes == 1, "Only one vote should succeed"
  assert Repo.aggregate(Vote, :count) == 1
end
```

## Related Files

- `lib/copi_web/live/player_live/show.ex` (lines 95-150)
- `priv/repo/migrations/20210816141921_create_votes.exs`
- `priv/repo/migrations/20260206100558_create_continue_votes.exs`
- `lib/copi/cornucopia/vote.ex`
- `lib/copi/cornucopia/continue_vote.ex`

prajakta128:
Resolves #2288

## Summary
Both `toggle_vote` and `toggle_continue_vote` used a check-then-act pattern (check if a vote exists, then separately insert or delete) with no database constraint backing it up. Concurrent requests could both pass the check and create duplicate votes.

`continue_votes` already had a unique constraint on `(player_id, game_id)`, but `votes` had none, and both handlers still relied on the racy in-memory check rather than the database.

## Changes
Scoped to two files:

- New migration: unique index on `votes(player_id, dealt_card_id)`
- `toggle_continue_vote`: replaced `Enum.find` + `delete!` with `Repo.delete_all` matching on the actual keys (handling `{n, _} when n > 0`, not just `{1, _}`), and replaced the plain `insert` with `on_conflict: :nothing, conflict_target: [:player_id, :game_id]`
- `handle_toggle_vote`: same pattern applied to card voting, using the new `votes` unique constraint as `conflict_target`

## Testing
Existing tests for both handlers continue to pass unchanged — their assertions check the resulting vote count/state, not the specific error path, so they remain valid under the new atomic-upsert behavior. Opening as a draft PR to let CI confirm before requesting review.

## Related
Two prior attempts (#2289, #2291) addressed this correctly in principle but were closed after months due to scope creep (rate limiting, an unrelated game-deletion handler, Python test file changes) and unresolved review comments (`coveralls.json` edits,
`delete_all` matching only `{1, _}`). This PR is scoped strictly to the migration and the two handlers.

@sydseter — you mentioned a migration script was already finished;
I wasn't able to locate it, so I've written one myself here. Happy to align with yours if it differs, just let me know.

### DIFF
--- a/copi.owasp.org/lib/copi_web/live/player_live/show.ex
+++ b/copi.owasp.org/lib/copi_web/live/player_live/show.ex
@@ -170,11 +170,20 @@ defmodule CopiWeb.PlayerLive.Show do
     else
       Logger.debug("Adding continue vote for player_id: #{player.id}, game_id: #{game.id}")
 
-      Copi.Repo.insert(
-        %Copi.Cornucopia.ContinueVote{player_id: player.id, game_id: game.id},
-        on_conflict: :nothing,
-        conflict_target: [:player_id, :game_id]
-      )
+      case Copi.Repo.insert(
+             %Copi.Cornucopia.ContinueVote{player_id: player.id, game_id: game.id},
+             on_conflict: :nothing,
+             conflict_target: [:player_id, :game_id]
+           ) do
+        {:ok, %{id: nil}} ->
+          Logger.debug("Continue vote already existed for player_id: #{player.id}, game_id: #{game.id}")
+
+        {:ok, _continue_vote} ->
+          Logger.debug("Continue vote added successfully for player_id: #{player.id}, game_id: #{game.id}")
+
+        {:error, changeset} ->
+          Logger.warning("Continue vote insert failed for player_id: #{inspect(player.id)}, game_id: #{inspect(game.id)}, errors: #{inspect(changeset.errors)}")
+      end
     end
 
     {:ok, updated_game} = game_module().find(game.id)
@@ -237,6 +246,9 @@ defmodule CopiWeb.PlayerLive.Show do
                    on_conflict: :nothing,
                    conflict_target: [:player_id, :dealt_card_id]
                  ) do
+              {:ok, %{id: nil}} ->
+                Logger.debug("Vote already existed for player_id: #{player.id}, dealt_card_id: #{card_id}, game_id: #{game.id}")
+
               {:ok, _vote} ->
                 Logger.debug("Vote added successfully for player_id: #{player.id}, dealt_card_id: #{card_id}, game_id: #{game.id}")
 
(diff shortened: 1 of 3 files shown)


sydseter:
@prajakta128 Please have a look at the comments

prajakta128:
@sydseter Thanks for the review! I've addressed all three Copilot comments:
- Migration now de-duplicates existing rows before creating the unique index, split into up/down since the delete isn't reversible
- Both toggle handlers now log whether the insert created a row, found an existing one, or errored, instead of ignoring the result
- Added a concurrency regression test using Task.async, matching the pattern from the original issue

Let me know if anything else needs adjusting.

sydseter:
@prajakta128 your code coverage is below the threshold.

PR #3410, opened 2026-08-26
Fix pointers

sydseter:
### Description

We have been using git lfs for some time, and it leads to a lot of issues when we change the files. The issue is that some of the files haven't been recently added to git lfs, so their pointers become inconsistent when working on those files after switching between branches, as the system expects these to be text pointers. This should take care of that.

### AI Tool Disclosure

- [x] My contribution does not include any AI-generated content
- [ ] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie etc.]`
    - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro etc.]`
    - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

### Affirmation

- [x] My code follows the [CONTRIBUTING.md](https://github.com/owasp/cornucopia/blob/master/CONTRIBUTING.md) guidelines


sydseter:
The image files aren‘t shown in the web preview due to the change in the git lfs index.

rewtd (approved):

PR #3455, opened 2026-09-06
Fix #2288: Prevent duplicate votes under concurrent requests

### LINKED ISSUE #2288: Race condition in voting system allows duplicate votes and data corruption

prajakta128:
## Description

Fixes #2288 by preventing duplicate votes caused by concurrent requests.

Previously, the vote handlers used a check-then-act approach: they first checked whether a vote existed and then inserted or deleted it. Under concurrent requests, multiple requests could pass the check before any vote was inserted, allowing duplicate vote records to be created.

This change moves duplicate prevention to the database and uses conflict-safe operations.

## Changes

- Added a unique database constraint on `votes(player_id, dealt_card_id)`.
- Added migration logic to remove existing duplicate vote records before creating the unique constraint.
- Added `Vote.changeset/2` with a unique constraint for `player_id` and `dealt_card_id`.
- Updated `toggle_vote` to:
  - use `Repo.delete_all/1` when removing a vote.
  - use `Repo.insert/2` with `on_conflict: :nothing` and `conflict_target: [:player_id, :dealt_card_id]` when adding a vote.
- Updated `toggle_continue_vote` to use conflict-safe inserts.
- Added unit tests for `Vote.changeset/2`.
- Added regression tests for concurrent vote and continue-vote requests.

## Why this fixes the issue

The previous implementation was vulnerable to a race condition because checking for an existing vote and inserting a new vote were separate operations.

The unique database constraint guarantees that only one vote can exist for the same player and dealt card, even when multiple requests are processed concurrently.

Using `on_conflict: :nothing` allows concurrent insert attempts to safely handle the conflict without creating duplicate records or raising an unnecessary database error.

## Testing

- Ran the complete test suite successfully.
- Result: **505 tests passed, 0 failures**.
- Verified the database migration successfully.
- Verified normal vote and un-vote behaviour.
- Added concurrency regression coverage to ensure concurrent requests cannot create duplicate votes.

## Related Issue

Closes #2288


sydseter:
The build is failing

Issue #3405, opened 2026-08-24
Create scenarios for the EoP game


sydseter:
Elevation of Privilege (EoP) is a serious tabletop card game designed by cybersecurity expert Adam Shostack in 2010 to make threat modeling accessible, engaging, and collaborative for software developers and architects. Instead of forcing engineering teams to sift through dry, academic security checklists, the game uses gamified mechanics to help players look at a system diagram and figure out "what can go wrong" before the code is actually built. 

We are looking for contributors to help us with the help pages for each of the EoP Cards https://cornucopia.owasp.org/edition/eop 

We need you to help us write a funny, but technical correct scenario with an example. 
You also need to provide a STRIDE analysis of the example.
We also need you to help players answer:

- What can go wrong?
- What are we going to do about it?

Remember to provide authoritative references (links) under "What can go wrong" and "What are we going to do about it?" that support your claims. The point is not to create an exhaustive list of everything that can go wrong, or that should be done, but to help the player start thinking about their own situation, and if you have a hilarious and absurd example, don't be afraid to use it (E.g: 2014 Steam "Blank Password" exploit)!
It‘s better if it is a hilarious example from the real world.

As an example, see: https://cornucopia.owasp.org/cards/AA2

To contribute, simply comment here and tell me which card you will do and wait for me to confirm.

To add your text, simply click on the "View source on GitHub" button on the bottom of the page of the card you want to do. See image.

<img width="777" height="422" alt="Image" src="https://github.com/user-attachments/assets/f955fb74-132b-4b4e-83fa-95828cc51d9b" />

khushal-winner:
Can I get to work on this? @sydseter

sydseter:
> Can I get to work on this? [@sydseter](https://github.com/sydseter)

Which card?

prakhar0x01:
Hi @sydseter ,

I'd like to work on TA2 — "An attacker can take advantage of your custom key exchange or integrity control which you built instead of using standard crypto"

Planning to use the 2013 Adobe password leak as the scenario example — where 153 million passwords were "encrypted" using ECB mode, making patterns visible in the ciphertext (someone even built a crossword puzzle from it). Will include the STRIDE analysis, "What can go wrong?" and "What are we going to do about it?" sections with authoritative references (OWASP, NIST).

PR #3489, opened 2026-09-12
docs(eop): add scenario for SP4 spoofing card

prajakta128:
### Problem

Card SP4 (Spoofing suit, EoP 5.0 edition) had an empty explanation.md template with no scenario or guidance. The threat: "An attacker can anonymously connect, because we expect authentication to be done at a higher level."

### Changes

- Added a scenario illustrating an internal API that trusts the gateway to have already authenticated requests
- Added a STRIDE analysis covering Spoofing and Elevation of Privilege
- Added "What can go wrong?" and "What are we going to do about it?" sections referencing NIST's Zero Trust Architecture guidance

### Testing

- Ran `npm run test` — full suite passes (250 tests)
- Verified the card renders correctly at /edition/eop/SP4/5.0/en on the local dev server

### Related

Addresses one card from #3405.

### DIFF
--- a/cornucopia.owasp.org/data/cards/eop-cards-5.0-en/spoofing/SP4/explanation.md
+++ b/cornucopia.owasp.org/data/cards/eop-cards-5.0-en/spoofing/SP4/explanation.md
@@ -1,11 +1,21 @@
-## Scenario: <name>'s ... scenario
+## Scenario: Meridian's internal API trusts the network, not the caller
 
 ### Example
+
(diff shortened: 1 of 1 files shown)


sydseter:
Thank you for this. It is great that you show interest in creating descriptions for the EoP cards. That is fantastic!

When reading your description. I feel like something is missing. Preferably, you should add something interesting and lighthearted (if not funny), to the description. It could be a real breach. 

If you want to keep it API centric, you can use the Equifax i 2017, or the 2021 an 2023 T-Mobile API breaches as an example.

https://en.wikipedia.org/wiki/2017_Equifax_data_breach
https://en.wikipedia.org/wiki/T-Mobile_data_breach

sydseter (approved):

sydseter:
Thank you for your contribution!

PR #3500, opened 2026-09-15
fix: sort and de-duplicate reference numbers before grouping

### LINKED ISSUE #3499: Card reference lists (CAPEC/ASVS/MASTG) render out of order and with duplicates when source YAML isn't pre-sorted
### Describe the bug
`group_number_ranges()` in `scripts/convert.py` compresses a list of reference numbers (CAPEC/ASVS/MASTG IDs) into ranges for the printed cards, e.g. `[25, 26, 27] -> "25-27"`. It does this using `itertools.groupby`, which only detects consecutive runs correctly if the input is already sorted ascending with no duplicates:

```python
data_numbers = [int(s) for s in data]   # no sort, no dedupe for k, g in groupby(enumerate(data_numbers), lambda x: x[0] - x[1]):  ...
```

Several `source/*mappings*.yaml` files list numbers out of order or with duplicates (they're hand-maintained), which produces incorrect, non-ascending, or duplicated text on the generated cards.

### To Reproduce
1. `source/mobileapp-mappings-2.0.yaml`, suit `CRM`, card `CRMX`, tag `capec: [20, 116, 117, 97, 112, 485]`
2. Run this through the converter's `check_make_list_into_text()` (used by `build_template_dict()` for every card tag)
3. Output is `"20, 116-117, 97, 112, 485"` — not ascending, so the reader can't tell at a glance that `112` and `485` haven't already been covered.

A worse case, `source/webapp-mappings-2.2.yaml` suit `WC`, card `JOB`: `capec: [184, 242, 416, 438, 441, 444, 523, 518, 519, 548, 636, 691]` renders as `"184, 242, 416, 438, 441, 444, 523, 518-519, 548, 636, 691"` — `518-519` appears *after* `523`, which reads as a typo/data error to anyone using the card for actual ASVS/CAPEC lookups.

A duplicate case: `["5", "5", "6"]` renders as `"5, 5-6"` instead of `"5-6"`.

### Expected behavior
Reference numbers should render sorted ascending, de-duplicated, with consecutive runs grouped into ranges, regardless of the order they appear in the source YAML.

### Affected files (confirmed by scanning all mapping files)
`mobileapp-mappings-1.0.yaml`, `mobileapp-mappings-1.1.yaml`, `mobileapp-mappings-2.0.yaml`, `companion-mappings-1.0.yaml`, `webapp-mappings-2.2.yaml`, `webapp-mappings-3.0.yaml` — dozens of individual card entries.

### Proposed fix
Sort and de-duplicate before grouping:
```python
data_numbers = sorted(set(int(s) for s in data))
```
Existing unit tests only ever exercise pre-sorted, de-duplicated input, which is why this hasn't been caught. Happy to submit a PR with the fix plus regression tests using the real unsorted data above.

prajakta128:
## What does this PR do?

Fixes an issue where CAPEC/ASVS/MASTG reference numbers could be rendered out of order or with duplicates when the source YAML mapping was not already sorted.

`group_number_ranges()` previously assumed that its input was sorted and de-duplicated before grouping consecutive numbers into ranges. Since the mapping files are hand-maintained, that assumption is not always valid.

This PR sorts and de-duplicates the numeric references before applying the existing range-grouping logic.

### Changes

* Sort reference numbers numerically before grouping.
* Remove duplicate reference numbers.
* Add regression tests for:

  * Unsorted mapping data.
  * Duplicate reference numbers.
* Update the function docstring to document the input assumption and behavior.

### Example

Before:

```text
[20, 116, 117, 97, 112, 485]
→ 20, 116-117, 97, 112, 485
```

After:

```text
[20, 116, 117, 97, 112, 485]
→ 20, 97, 112, 116-117, 485
```

Duplicate input such as:

```text
[5, 5, 6]
```

now correctly renders as:

```text
5-6
```

### Testing

* `python -m pytest tests/scripts/convert_utest.py -v`
* **156 tests passed**
* 0 failures

Fixes #3499


prajakta128:
Hi @sydseter @rewtd — noticed the ClusterFuzzLite PR fuzzing / PR (address) check is failing, wanted to flag it before review rather than leave it silent.

The crash is in capec_map_enricher_fuzzer, not in the code this PR touches (group_number_ranges() / check_make_list_into_text() in convert.py). The failure itself is:

Exception ignored in: <function _removeHandlerRef at ...>
TypeError: 'NoneType' object is not callable

That's happening inside Python's own logging module during interpreter shutdown, not in application code, and none of my changed functions are reachable from that fuzz target. The job log also shows ClusterFuzzLite couldn't download any previous build artifact to confirm whether this crash pre-dates this PR ("Could not download artifact... Build ... does not exist"), so it wasn't able to verify it as a regression from this change.

Is this a known/pre-existing flake on capec_map_enricher_fuzzer, or should I look into it separately? Happy to open a separate issue for it if it's new — just didn't want to bundle an unrelated fix into this PR.

sydseter:
We don't always want the dash if there are number ranges that followes in a sequence. It's only the attribute that has the "_print" postfix where we want to use the dash. For all other attributes, they should not have it. This is because the identifiers are used in linking to external pages.

sydseter:
Thank you for paying attention. Bug reported.

sydseter:
Please be aware of the contributor guidelines: https://github.com/OWASP/cornucopia/blob/master/CONTRIBUTING.md
You should use our default pull-request template when creating pull-request.

I will close this pull-request, but feel free to open a new pull-request in the correct way.

